PR #9, opened 2018-08-22
move proton from python2.7 to python3

swick:
python2 is EOL in 2020. It's not even packaged in the freedesktop sdk. There is no reason for this to be python2.

Patch makes Proton work in Steam from flatpak. It's not thoroughly tested.

### DIFF
--- a/proton
+++ b/proton
@@ -1,4 +1,4 @@
-#!/usr/bin/python2.7
+#!/usr/bin/python3
 
 #script to launch Wine with the correct environment
 
@@ -309,7 +309,7 @@ if "nod3d11" in config_opts:
         del dlloverrides["dxgi"]
 
 s = ""
-for dll, setting in dlloverrides.iteritems():
+for dll, setting in dlloverrides.items():
     if len(s) > 0:
         s = s + ";" + dll + "=" + setting
     else:
@@ -376,7 +376,7 @@ def dump_dbg_scripts():
         f.write("cd \"" + os.getcwd() + "\"\n")
         dump_dbg_env(f)
         f.write("\t\"" + wine_path + "\" winedbg \"$@\"\n")
-    os.chmod("/tmp/proton_winedbg", 0755)
+    os.chmod("/tmp/proton_winedbg", 0o755)
 
     with open("/tmp/proton_winedbg_run", "w") as f:
         f.write("#!/bin/bash\n")
@@ -393,7 +393,7 @@ def dump_dbg_scripts():
         f.write(")\n")
         dump_dbg_env(f)
         f.write("\t\"" + wine_path + "\" winedbg \"${@:-${DEF_CMD[@]}}\"\n")
-    os.chmod("/tmp/proton_winedbg_run", 0755)
+    os.chmod("/tmp/proton_winedbg_run", 0o755)
 
     with open("/tmp/proton_gdb_attach", "w") as f:
         f.write("#!/bin/bash\n")
@@ -408,7 +408,7 @@ def dump_dbg_scripts():
         f.write("WPID_DEC=$(printf %d 0x$WPID_HEX)\n")
         dump_dbg_env(f)
         f.write("\t\"" + wine_path + "\" winedbg --gdb $WPID_DEC\n")
-    os.chmod("/tmp/proton_gdb_attach", 0755)
+    os.chmod("/tmp/proton_gdb_attach", 0o755)
 
     with open("/tmp/proton_gdb_run", "w") as f:
         f.write("#!/bin/bash\n")
@@ -425,7 +425,7 @@ def dump_dbg_scripts():
         f.write(")\n")
         dump_dbg_env(f)
         f.write("\t\"" + wine_path + "\" winedbg --gdb \"${@:-${DEF_CMD[@]}}\"\n")
-    os.chmod("/tmp/proton_gdb_run", 0755)
+    os.chmod("/tmp/proton_gdb_run", 0o755)
 
     with open("/tmp/proton_run", "w") as f:
         f.write("#!/bin/bash\n")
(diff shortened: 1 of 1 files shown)


valeth:
@swick the `user_settings.sample.py` file should also be updated.
It's still referring to python2.7, which could lead to future issues if people start to use is as a basis for their configuration.

valeth:
`vrclient_x64/gen_wrapper.py` is another python2 file
There might be more spread across the entire repository.

tcarrio:
```bash
$ find . -type f -exec grep -e "python2" {} /dev/null \;
./user_settings.sample.py:#!/usr/bin/python2.7
./lsteamclient/gen_wrapper.py:#!/usr/bin/python2
./vrclient_x64/gen_wrapper.py:#!/usr/bin/python2
./ffmpeg/tools/normalize.py:#!/usr/bin/env python2
./ffmpeg/tools/zmqshell.py:#!/usr/bin/env python2
./proton:#!/usr/bin/python2.7
```

Of these it seems that the `lsteamclient` and `vrclient_x64` directories are maintained here along with `user_settings.sample.py` and `proton` (..duh). The `ffmpeg` tools are submodules. So those 4 need to be updated to support Python 3. 

There should be confirmation that the ffmpeg tools aren't used in any way, since without a Python 2 interpreter this may cause unexpected behavior.

valentindavid:
Could you use `/usr/bin/env python3` instead?

MayeulC:
Instead of forcing either python 2 or 3, it would be best to make the code compatible with both, and change the shebang to `/usr/bin/python`, or better, `/usr/bin/env python`. Also stated in https://github.com/ValveSoftware/Proton/issues/15#issuecomment-414907327

Though python 3 compatibility should be "good enough".

valentindavid:
@MayeulC, The correct meaning of just `python` is Python 2. So no, `python3` should be used.

MayeulC:
@valentindavid: There is no standard for this. On arch Linux it's python3. If the code is platform-agnostic, it shouldn't matter anyway. See for example https://www.eflglobal.com/version-agnostic-python-part-1/

But if a hard requirement on python3 is acceptable (which is probably the case, at least more so than a hard python2 requirement), then I'm fine with the way this patch is written.

proton
```diff
@@ -469,11 +469,11 @@ elif sys.argv[1] == "waitforexitandrun":
 elif sys.argv[1] == "getcompatpath":
     #linux -> windows path
     path = subprocess.check_output([wine_path, "winepath", "-w", sys.argv[2]], env=env, stderr=lfile)
-    sys.stdout.write(path)
+    sys.stdout.write(path.decode('utf-8'))
```
MayeulC, line 472: I don't think this is necessary.

valentindavid:
Docs say that recommended shebang is `#!/usr/bin/env python3`
https://docs.python.org/3/using/unix.html

For python 2.7 they doc talks about `python`.
https://docs.python.org/2.7/using/unix.html

That is the standard.

adds68:
When will this get merged?

DelusionalLogic:
@MayeulC That's just wrong.

The python command is standardized in [pep-0394](https://www.python.org/dev/peps/pep-0394/), as a response to Arch Linux switching `/usr/bin/python` to `python3`. According to the standard new scripts should use `#!/usr/bin/python` _only_ if it's source compatible with both. Any script only compatible with one version should instead use the version specific shebang.

Please keep in mind that 0394 is a moving target.

proton
nanonyme: Yes, it is. In Python3 stdout eats unicode. If you want to write bytes, you need stdout.buffer

nanonyme:
I'm strongly in favour of using /usr/bin/env instead of hardcoding Python path. It makes less assumptions about distro prefix

proton
nanonyme: I actually recommend using buffer instead since then you don't need to make assumptions about encoding

MayeulC: That's [not](http://www.macfreek.nl/memory/Encoding_of_Python_stdout) [quite](https://stackoverflow.com/a/15740694/3795597) true, as stdout ingests unicode strings, but it could sometimes misdetect the terminal's encoding. Of course, forcing utf-8 will make things more predictable, but might also have some unintended consequences on non-utf-8 terminals (rare on Linux).
Please note that `sys.stdout.write("éè€\u2611😼")` works for me. Anyway, that's not really the most pressing issue here, both choices are fine, but it would be better to stick to it for consistency.

nanonyme: I've been using sys.stdout.buffer.write(bytedata) where the string came from another OS with different encoding. It doesn't do any guessing but just passthroughs, similar to the Python2 code

nanonyme: In Python3 yours was unicode. Try with b"foo" instead

nanonyme:
Created #96 as counter-proposal so we can delay the Python3 migration a bit still

proton
MayeulC: All right, I realized I misunderstood what you were saying, my bad. Feel feel to write a review suggesting the use of  `buffer` instead; I'll delete this review later :)

aeikum:
Unfortunately Apple doesn't ship python3 on macOS yet, so we need to maintain python2.7 compatibility.

valentindavid:
@aeikum Have they planned to *not* remove python2 on Mac? I would not be surprised if they just remove Python all together. You might want to embed python3 for Mac.

DelusionalLogic:
@aeikum The scripts seem fairly trivial. Is there any reason they couldn't be made source compatible?

proton
valentindavid: What about using `universal_newlines=True` to `subprocess.check_output`. Then we get it compatible python 2 and 3.

aeikum:
@DelusionalLogic I didn't mean to imply it can't be, just explaining that we can't drop python2.7 support.

nanonyme:
@DelusionalLogic as pointed out by @tcarrio earlier, this project has ffmpeg dependency and Python 2 is used there too. We're going to solve this in Flatpak Steam by bundling Python 2 for now, we just needed Proton to support having Python in /app which is now done through separate PR. This allows doing Python 3 migration without any specific rush.

proton
nanonyme: That will invoke automatic decoding in Python 3 and will totally not be the same as the Python 2 code currently is. The Python 2 code currently simply doesn't care about the encoding and passes through output to console. The actually equivalent code means two separate code paths where on Python 2 you write to sys.stdout and on Python 3 you write to sys.stdout.buffer. I've done this in some codebases that needed to support both and it's the only idiot proof way to do it

LucyDemooon:
proton not work an i have only python 3

PR #20, opened 2018-08-22
Runtime config suggestion for specific game options

tcarrio:
Updated README to reflect specific-game environment variable modification alongside the existing global option available through `user_settings.py`.

I didn't see a contribution guideline and don't know whether you take PRs here.

I'm introducing this as a description of how users can set variables on a specific game as well as globally. The README does a good job of showing how proton can be modified for all games. 

An idea for `user_settings.py` in the future, and something I might look into:

Allow specific games to have their environment variables modified through this file. It could be done by providing the unique ID for the game as a separate nested dictionary. Just an idea :smile:

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,8 +170,11 @@ Runtime Config Options
 Proton can be tuned at runtime to help certain games run. The Steam client sets
 some options for known games using the <tt>STEAM_COMPAT_CONFIG</tt> variable.
 You can override these options using the environment variables described below.
-The best way to set these environment overrides is by renaming
+The best way to set these environment overrides for all games is by renaming
 "user_settings.sample.py" to "user_settings.py" and modifying it appropriately.
+If you want to set a specific game to a different setting than the defaults, 
+use the `Set Launch Options` under the games `Properties`. You can launch the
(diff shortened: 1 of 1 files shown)


jwhite66:
The first part of this change is clearly correct and a nice clarification.
The next part is not entirely true; the PROTON_XXX environment variables will be picked up that way, but WINEDEBUG is not.  It may be that the correct fix is to adjust the proton script until your change is accurate; we need to review that.

tcarrio:
Yeah, it seems there could be checks against whether the key has been set in `env` already (since it receives all of `os.environ`). 

A small helper function may even suffice here so behaviour could be set in one location?

tcarrio:
This may also be the desired behaviour. Overriding the systematically decided ones doesn't seem particularly necessary- You would break the script overriding certain WINE variables, for example. But then, why would you? 

Allowing existing environment variables to supercede the scripts could be the right choice. Let me know after you review it :smile:

tcarrio:
For now, I'm specifying the PROTON_ variables specifically in the README. The option to allow other environment variables can be visited again.

Issue #27, opened 2018-08-22
build_proton.sh uses 4 jobs no matter how many cores are available


tcarrio:
The build process could be sped up on systems with more than 4 cores available. 

Right now it sets `JOBS=-j5` which is a reasonable default but doesn't optimize based on available resources.

Mikerr1111:
Yeah, I think that will eventually change but for now, 4 cores works for a vast majority of people.

Valmar33:
It would be better to just use something like `make -j$(nproc)`

tcarrio:
@Valmar33 I was initially going to leave it as that, however I'm unfamiliar whether macOS includes this command by default and what the behavior would be on that platform, or any others that don't.

If macOS includes this, that would be a better move. I simply wanted to support Linux/macOS building and maintain the same `core_count + 1` as it was previously.

flukejones:
On macOS the command `sysctl -n hw.ncpu` should work. `nproc` is not available.

tcarrio:
@Luke-Nukem Awesome, this seems like this should work! :confetti_ball:

2A4U:
Compile based on number of threads available on system (not cores):

readarray -t -s3 -n3 threads <<< $(lscpu) ; export threads=$((${threads[0]/*:}*${threads[2]/*:}))

or

readarray -t -s3 -n3 threads <<< $(lscpu) ; export threads=$((${threads[0]/*:*[[:space:]]}*${threads[2]/*:*[[:space:]]}))

then:

make -j${threads}

pchome:
`build_proton.sh -j5`

aeikum:
Fixed by 0d9ec29b5e3dea0d536e9da230f4f9f97d03a9fc.

PR #28, opened 2018-08-22
Optimize build_proton.sh by setting jobs to available cores

tcarrio:
build_proton uses available cores for job count. Defaults to 5.

This uses a combination of `nproc` (very common Linux command), and `systctl` (standard hw manager on macOS) as well a default value. 

This was tested on Ubuntu 18.04 with both `bash` and `sh`.

### DIFF
--- a/build_proton.sh
+++ b/build_proton.sh
@@ -16,7 +16,7 @@ usage()
 
 set -e
 
-JOBS=-j$(expr `nproc 2>/dev/null||sysctl -n hw.ncpu 2>/dev/null||echo 4` + 1)
+JOBS=-j"$(( $(nproc 2>/dev/null||sysctl -n hw.ncpu 2>/dev/null||echo 4) + 1))"
 PLATFORM=$(uname)
 
 [ -z "$STEAM_RUNTIME" ] && STEAM_RUNTIME="$HOME/steam-runtime"
--- a/build_proton.sh
+++ b/build_proton.sh
@@ -16,7 +16,7 @@ usage()
 
 set -e
 
-JOBS=-j$(expr `nproc 2>/dev/null||echo 4` + 1)
+JOBS=-j$(expr `nproc 2>/dev/null||sysctl -n hw.ncpu 2>/dev/null||echo 4` + 1)
 PLATFORM=$(uname)
 
 [ -z "$STEAM_RUNTIME" ] && STEAM_RUNTIME="$HOME/steam-runtime"
--- a/build_proton.sh
+++ b/build_proton.sh
@@ -16,7 +16,7 @@ usage()
 
 set -e
 
-JOBS=-j5
+JOBS=-j$(expr `nproc 2>/dev/null||echo 4` + 1)
 PLATFORM=$(uname)
 
 [ -z "$STEAM_RUNTIME" ] && STEAM_RUNTIME="$HOME/steam-runtime"


tcarrio:
This fixes issue #27

build_proton.sh
```diff
@@ -16,7 +16,7 @@ usage()
 
 set -e
 
-JOBS=-j5
+JOBS=-j$(expr `nproc 2>/dev/null||sysctl -n hw.ncpu 2>/dev/null||echo 4` + 1)
```
DirectorX: I do not think it's compatible with Mac

masom:
```
> expr `nproc 2>/dev/null||sysctl -n hw.ncpu 2>/dev/null||echo 4` + 1
> 9
```

Worked on OSX 10.13

DirectorX: `nproc` or `sysctl` ?

tcarrio: nproc is standard in Linux
sysctl is standard on macOS
Did this not work for you?

masom: `sysctl` but the way the expression evaluation is setup, if nproc is missing it will fallback to `sysctl` on OSX.

DirectorX: I don't have Mac and also missed the note about this on the PR description.

tcarrio: The command is written to cascade down until success. You wont have nproc on macOS but that's fine since the OR checks the command status and moves on to sysctl. Same for that. If neither are supported on your platform it will default to 5.

tcarrio: I got beat thanks @masom :+1:

donovan:
Unless there are compatibility reasons not to can I suggest writing this using modern bash? The current code is using some deprecated features (see shellcheck output below):

```
JOBS=-j$(expr `nproc 2>/dev/null||sysctl -n hw.ncpu 2>/dev/null||echo 4` + 1)
         ^-- SC2003: expr is antiquated. Consider rewriting this using $((..)), ${} or [[ ]].
              ^-- SC2046: Quote this to prevent word splitting.
              ^-- SC2006: Use $(..) instead of legacy `..`.
```

like this:

`JOBS=-j"$(( $( nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 4 ) + 1 ))"`

tcarrio:
I don't see an issue with that. I'll go ahead and update it

aeikum:
Thanks, I'll test and merge this in soon.

pchome:
Here is the small example to quick-test this, and/or extend to allow user pick `-j` value or even pass more parameters to `make` command:

`$ ./1.sh -j2 -- -j3 --quiet`
```sh
./1.sh: -j2 -- -j3 --quiet
./1.sh: jobs=-j2
./2.sh: -j3 --quiet
./2.sh: jobs=-j3
```
Code:
`1.sh`
```sh
#!/bin/sh

. ./common.sh
echo "$0: jobs=$JOBS"

./2.sh $@
# make $@
```
`2.sh`
```sh
#!/bin/sh

. ./common.sh
echo "$0: jobs=$JOBS"
```
`common.sh`
```sh
[[ -n $@ ]] && echo $0: $@ || echo "empty"

JOBS=-j"$(( $( nproc 2>/dev/null || sysctl -n hw.ncpu 2>/dev/null || echo 4 ) + 1 ))"

while [[ -n $1 ]]; do
  case "$1" in
    -j[0-9]) JOBS="${1}" ; shift ;;
         --) shift ; break ;;
          *) shift ;;
  esac
done
```

aeikum:
Thanks, this was merged as 0d9ec29b5e3dea0d536e9da230f4f9f97d03a9fc.

PR #96, opened 2018-08-22
Hashbang sanitizing

nanonyme:
Use /usr/bin/env in hashbangs to allow different prefixes to be used

### DIFF
--- a/lsteamclient/gen_wrapper.py
+++ b/lsteamclient/gen_wrapper.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python2
+#!/usr/bin/env python2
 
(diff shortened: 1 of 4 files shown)


aeikum:
Thanks.

nanonyme:
Can you please ping us when this is released so we can revert https://github.com/flathub/com.valvesoftware.Steam/commit/1d0eb18e76da2b3712a68748ddebdebaa8185024? We would prefer not mangling game data if at all possible.

aeikum:
@nanonyme I believe this is currently live, can you restart your Steam client to be sure Proton is up to date and double check on your end?

nanonyme:
Thanks, I'll handle

Issue #117, opened 2018-08-22
cannot compile if ccache is installed on the host system


compholio:
The build uses ccache if it's installed, but it only checks that ccache is installed on the host system.  If ccache is not also installed within the chroot environments then you will see (a version of) the following error:
```
checking for gcc... /usr/bin/ccache gcc
checking whether the C compiler works... no
configure: error: in `/home/ehoover/src/proton/build/wine.win64':
configure: error: C compiler cannot create executables
```
You can work around this by running the following commands:
```
schroot --chroot steamrt_scout_beta_i386 -- sudo apt-get install ccache
schroot --chroot steamrt_scout_beta_amd64 -- sudo apt-get install ccache
```
The easiest way to deal with this is probably to check for the ccache binary existence in build_proton.sh, for example:
```
if [ "$($I386_WRAPPER ls $CCACHE)" = "" ]; then CCACHE=""; fi
if [ "$($AMD64_WRAPPER ls $CCACHE)" = "" ]; then CCACHE=""; fi
```

PR #128, opened 2018-08-22
fix compiling when ccache is not available in the chroot environment

compholio:
Fixes Issue #117

### DIFF
--- a/build_proton.sh
+++ b/build_proton.sh
@@ -566,6 +566,9 @@ else
     fi
 fi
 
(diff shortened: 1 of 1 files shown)


deah77:
U cant help me  please

aeikum:
The build system has been totally reworked since this. I'd love to get ccache working again, but it's tricky with multiple build VMs sharing a cache dir. I think newer ccaches might work better. Patches welcome...

PR #237, opened 2018-08-23
build_proton: Fix build wine32 typo

Gravemind:
`./build_proton.sh --build wine32` builds wine64 instead of wine32, looks like a typo.


aeikum:
Thanks!

PR #241, opened 2018-08-23
Add Python 2 requirements in PREREQS.md

EzoeRyou:
Proton(and git submodule ffmpeg) depends on deprecated and soon-to-be-out-supported Python 2. If the system doesn't have Python 2 installed, Steam Play fails without any diagnostic massage.

Until Proton migrate to Python 3, PREREQS.md should have mentioned the Python 2 requirements.

PR #261, opened 2018-08-23
Make the scripts work regardless of the python version

### LINKED ISSUE #15: Games do not run without Python 2.7
I attempted to run some games using Proton (including Quake 3 and Audiosurf, both of which are rated "Platinum" on WineHQ and run normally under normal Wine), and they do not run at all. After clicking "Play", steam will show in the friends list that I am playing the game for a fraction of a second, and then it will immediately change back to "Online". The games do not show up at all, with no windows appearing. I tested using Ubuntu 18.04.1 LTS, with a Radeon R9 270X GPU, and the latest mesa drivers installed from the PPA shown in the guide. I ran Steam via the terminal in order to check if there were any errors, and I saw no errors at all after running the games.
UPDATE: It turns out that after installing the python2.7 package, Proton functions properly. It seems that on some installs, Python 2.7 is not present, which causes Proton to not function.

MayeulC:
This builds upon, thus closes #9, which I rebased on upstream, and obsoletes #241. It also resolves #15.

This PR should work in theory, but please not that **I did not test it**. It would be great to setup CI with this repository.

I should also mention that I submitted similar patches to upstream ffmpeg. Unfortunately I cannot test the patches right now. Hopefully someone will chime in and do just that.

### DIFF
--- a/proton
+++ b/proton
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 
 #script to launch Wine with the correct environment
 
@@ -309,7 +309,8 @@ if "nod3d11" in config_opts:
         del dlloverrides["dxgi"]
 
 s = ""
-for dll, setting in dlloverrides.items():
+for dll in dlloverrides:
+    setting = dlloverrides[dll]
     if len(s) > 0:
         s = s + ";" + dll + "=" + setting
     else:
@@ -457,6 +458,13 @@ def run():
         else:
             run_wine([wine_path] + sys.argv[2:])
 
+if sys.version_info[0] == 2:
+    binary_stdout = sys.stdout
+elif sys.version_info[0] == 3:
+    binary_stdout = sys.stdout.buffer
+else:
+    raise Exception("Unsuported python version")
+
 #determine mode
 if sys.argv[1] == "run":
     #start target app
@@ -469,11 +477,11 @@ elif sys.argv[1] == "waitforexitandrun":
 elif sys.argv[1] == "getcompatpath":
     #linux -> windows path
     path = subprocess.check_output([wine_path, "winepath", "-w", sys.argv[2]], env=env, stderr=lfile)
-    sys.stdout.write(path.decode('utf-8'))
+    binary_stdout.write(path)
 elif sys.argv[1] == "getnativepath":
     #windows -> linux path
     path = subprocess.check_output([wine_path, "winepath", sys.argv[2]], env=env, stderr=lfile)
-    sys.stdout.write(path.decode('utf-8'))
+    binary_stdout.write(path)
 else:
     log("Need a verb.")
     sys.exit(1)
--- a/proton
+++ b/proton
@@ -2,6 +2,8 @@
 
 #script to launch Wine with the correct environment
 
+from __future__ import print_function
+
 import filecmp
 import json
 import os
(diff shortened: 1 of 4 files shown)


proton
```diff
@@ -469,11 +470,11 @@ elif sys.argv[1] == "waitforexitandrun":
 elif sys.argv[1] == "getcompatpath":
     #linux -> windows path
     path = subprocess.check_output([wine_path, "winepath", "-w", sys.argv[2]], env=env, stderr=lfile)
-    sys.stdout.write(path.decode('utf-8'))
```
MayeulC: Please note that this was done to maximize compatibility with as few changes as necessary. @nanonyme expressed some objections to this change in https://github.com/ValveSoftware/Proton/pull/9#discussion_r211950495

The same change also happens four lines below.

MayeulC: Please note that this was done to maximize compatibility with as few changes as necessary. @nanonyme expressed some objections to this change in https://github.com/ValveSoftware/Proton/pull/9#discussion_r211950495

The same change also happens four lines below.

nanonyme: Yes, it will only work in Python 2. Python 3 subprocess outputs bytes and stdin takes unicode. It will raise an exception as implicit decoding is no longer allowed

MayeulC: mmm.. reverting this change actually doesn't seem to have any adverse effects

proton
```diff
@@ -457,6 +458,13 @@ def run():
         else:
             run_wine([wine_path] + sys.argv[2:])
 
+if sys.version_info[0] == 2:
```
MayeulC, line 463: @nanonyme I finally bit the bullet and wrote this to mux the binary input to the correct binary output, according to your suggestion in https://github.com/ValveSoftware/Proton/pull/9#discussion_r212055974

I would have preferred to avoid explicitly different codepaths if it had peen possible.

Technically, it is possible to leave it as `sys.stdout.write(path.encode('utf-8'))`, but would cause weird issues on non utf8-encoded outputs that subprocess might spit out (or just outright fail).

nanonyme: Also decoding is wrong since Python 2 expects bytes and decode results in unicode so you get implit encode inside write

MayeulC: I meant `decode`, sorry. `encode` wouldn't work. But yeah, this would rely on implicit behaviour. Are you good with the changes now?

nanonyme: This looks fine to me

nanonyme: Note you can also
from __future__ import print_function

MayeulC: Mmm, good point, that could remind people to keep the code compatible... On the other hand, there are very few `print` statements. I'd rather see a CI system that checks both with python 2 and 3.

nanonyme: It's just that parenthesis does different things with print statement and function which can be confusing

aeikum:
@MayeulC @nanonyme Please let me know when you think this is ready to be merged and I'll review and test it. Thanks.

nanonyme:
I was just nitpicking with the print_function part, I'm pretty much done with comments.

MayeulC:
@nanonyme, I agreed with your last point, and I am going to add a commit to address this in a few minutes. @aeikum, it should be ready after that.

Edit: I gave credit to @nanonyme in the commit message. Please tell me if that's not OK.
I also didn't bother including this line in the `user_settings.sample.py` as I highly doubt this would be useful.

nanonyme (approved):

aeikum:
Thanks, this worked in both python2 and 3 for me.  This has been merged.

PR #266, opened 2018-08-23
Create issue_template.md for new issue field population

asdf1234-deleted:


kisak-valve:
Hello @mtdeguzis.

It may be more valuable to have something along the lines of

Proton version:
GPU model and driver (e.g. AMD RX480 / mesa 18.1.6 and llvm 7.0):
System information from Steam (`Steam` -> `Help` -> `System Information`) in a [gist](https://gist.github.com/):
Checked for system updates: [Y/N]
If this is a compatibility report, checked if an issue report for the game does not already exist: [Y/N]

asdf1234-deleted:
np, I will update the PR soon with your suggestions.

castrojo:
Probably wouldn't hurt to list the minimum versions of the drivers in the template? That might cut down on people reporting bugs on from older drivers.

issue_template.md
```diff
@@ -0,0 +1,20 @@
...
+
+* Steam client version:
+* Distro / OS version (e.g. Arch Linux 64-bit): 
+* CPU vendor and model (e.g. Intel i7-7700HQ):
+* Graphics card/chip vendor and model (e.g. NVIDIA GTX 1060): 
+* Graphics driver version (e.g. Nvidia 396.55): 
+* Opted into Steam client beta?: 
+* Have you checked for system updates for your distro/OS?:
+* If this is in reference to a game, what game?:
+* Have you reviewed the [community reports](https://github.com/ValveSoftware/Proton/wiki/Community-Compatibility-Reports) on the wiki?:
```
Elkasitu, line 11: Would be useful to link to the PREREQS.md as well IMO

ghost (approved):

ghost:
Edit: First of all, this is a great idea.

Currently the created issues can be categorized as:
1. Bug reports
2. Feature requests
3. Whitelist requests

The template doesn't cover whitelist requests which make a big part of all issues. In case of feature requests the template does also seem not to be fitting completely in my opinion. For this reason I suggest to modify the template to suit all three categories and tell people to remove specific lines in specific cases.

When I look at the Google Docs compatibility table, I have seen entries like those for Age of Empires 2 HD who state that this game would run well with Proton. But to achieve this they have renamed the game file to Launcher.exe because the actual launcher is broken.
The purpose of Proton is that you just click on Install and then run the game. As a result the statement that this game would run well with Proton is simply wrong. Instead, the right way to go would be creating a bug report to fix the issues with the launcher in order to run the game.

To prevent wrong requests it would be helpful when people were asked the following questions when they like to submit a whitelist request:
1. How long have you played the game?
2. Have you experienced any issue during this time?
3. Have you done anything else than clicking "Install" in your library?

Just to make sure that people who haven't understood what Proton is for do not request malfunctioning software to be whitelisted.

issue_template.md
```diff
@@ -0,0 +1,20 @@
+### Your system information
+
+* Steam client version:
+* Distro / OS version (e.g. Arch Linux 64-bit): 
+* CPU vendor and model (e.g. Intel i7-7700HQ):
+* Graphics card/chip vendor and model (e.g. NVIDIA GTX 1060): 
+* Graphics driver version (e.g. Nvidia 396.55):
```
ghost, line 7: Line 7 might be considered a problem by multiple groups of people within the open source community in the current state. Because of several reasons:

- Valve is developing open source drivers with Mesa
- This proprietary software (Edit: That is currently mentioned as example) is not complying to values that open source software represents at all.
- The company developing this proprietary software could contribute to the Nouveau project, just like AMD and Intel do support their respective open source projects with Mesa, but they refuse.

For these reasons it is not reasonably considerable to promote this proprietary software in this open source project by naming it directly as long as we have our own open source community drivers that replace it.

Elkasitu: I hate nvidia as much as the next guy, but if using these proprietary nvidia drivers makes people completely move to linux and delete windows forever because proton works for them, then I think it's a worthy trade-off

ghost: Well, my comment wasn't meant as blocker but as an info. A Mesa version would cover all GPUs and Mesa provides the standard, out of the box drivers for all graphics APIs that are common for open source software. Installing proprietary drivers is an option but not the default. And it is never something we like to see on Linux, because beside intellectual property which can be protected by putting small bits of code in firmware blobs (if need be...), there seem to be no valid reasons for someone who sells __hardware__ to hide everything other than shady and suspicious ones. 

By adding the LLVM version in the example you also remind people who have a non standard Mesa version to tell which compiler was used to assign issues better to projects. But I would suggest to mark this entry as optional because when you install Ubuntu 18.04 as an example and don't fiddle, it's clear how the driver has been built and you shouldn't be bothered with LLVM and stuff that might confuse you.

When people use the proprietary drivers it should __not__ be necessary to show them how their version string looks like - everyone who is able to type numbers should know how a version of a program looks like. The danger that someone misses to tell anything relevant does only exist when the driver can be compiled by third parties because it's open source.

steelbrain (commented):
Thoughts on having multiple issue templates where users can select if it's a whitelist request or a bug report?

Give this a read https://blog.github.com/2018-01-25-multiple-issue-and-pull-request-templates/

asdf1234-deleted:
Dude that rocks! Multiple templates would be slick.

asdf1234-deleted:
Closing in favor of #266

Issue #121, opened 2018-08-22
[Mac OS] Corresponding binaries of glslang


3t0n:
Linux binaries of glslang can't be run on Mac OS. 
So should be replaced on Mac (https://github.com/KhronosGroup/glslang/releases)

zakk4223:
MoltenVK actually pulls in glslang as part of its build; the build script could just swap that linux binary with a symlink to the MoltenVK glslangvalidator

flukejones:
See issue #29 also - duplicate.

Issue #29, opened 2018-08-22
glslang should be a git submodule


flukejones:
I notice that glslangValidator is precompiled - this should probably be a git submodule to allow for easier building on macOS (and other Linux distros).

cjwijtmans:
i agree. i was wondering why this is not a submodule.

peterson432:
yea, glslang validator is pre-compiled and it creates a difficulty for building an mac OS and linux, it should be a submodule for easier access.

suprstarrd:
Just a note, two years in the future: macOS is *not* a Linux distro. It is a Unix-based operating system, while Linux is a Unix-based *kernel*. macOS does not run Linux, it runs *Unix*.

flukejones:
> Just a note, two years in the future: macOS is _not_ a Linux distro. It is a Unix-based operating system, while Linux is a Unix-based _kernel_. macOS does not run Linux, it runs _Unix_.

What was the point of this reply other than to be correct? The initial issue post was a mistype and everyone else seemed to get the point.

cjwijtmans:
> Just a note, two years in the future: macOS is _not_ a Linux distro. It is a Unix-based operating system, while Linux is a Unix-based _kernel_. macOS does not run Linux, it runs _Unix_.

(L)inux (I)s (N)ot (U)ni(X)

flukejones:
Christ... 
*unsubscribe*

suprstarrd:
The point was so that people could be informed, I'm sure someone out there probably might assume that. Yeah, likely a good idea to unsubscribe. Sorry!

cjwijtmans:
> The point was so that people could be informed, I'm sure someone out there probably might assume that. Yeah, likely a good idea to unsubscribe. Sorry!

Except you are totally incorrect but ok.

Issue #2, opened 2018-08-21
libjpeg-turbo dependency for macOS


spotlightishere:
libjpeg-turbo depends on `nasm` from Homebrew/MacPorts, as it's not natively provided by the Xcode toolchain.

flukejones:
Other dependencies for macOS seem to be:

- `realpath`
- `i686-w64-mingw32-gcc`
- `x86_64-w64-mingw32-gcc`
- `meson` and `ninja`

henrebotha:
`meson` depends on `ninja` so just `brew install meson` will take care of the latter.

You can get `realpath` using `brew install coreutils`.

I'm no C developer so I'm not sure how to grab the other two.

zsxsoft:
``autoconf`` and ``automake`` is required too

petmac:
`brew install mingw-w64` should work for the mingw dependencies.

henrebotha:
My build fails on:

```
FAILED: src/dxvk/src@dxvk@@dxvk@sta/dxvk_clear_buffer_u.h
/Users/henrebotha/dev/proton/glslang/bin/glslangValidator -V --vn dxvk_clear_buffer_u ../../dxvk/src/dxvk/shaders/dxvk_clear_buffer_u.comp -o src/dxvk/src@dxvk@@dxvk@sta/dxvk_clear_buffer_u.h
/bin/sh: /Users/henrebotha/dev/proton/glslang/bin/glslangValidator: cannot execute binary file
```

The included `glslangValidator` binary is compiled for GNU/Linux. I'm now doing `brew install glslang` & I'll try and use that instead.

petmac:
I deleted that `glslangValidator` locally and the build seemed to continue. Not a proper solution obviously, but time was a factor.

3t0n:
Change glslangValidator with proper binaries from https://github.com/KhronosGroup/glslang/releases
Vote for https://github.com/ValveSoftware/Proton/issues/29 to make submodule

Also, you can download Proton3.7 dist for MacOS High Sierra [here](https://github.com/3t0n/Proton/releases/tag/20180822)

aaronfranke:
macOS support has been removed: https://github.com/ValveSoftware/Proton/commit/a84120449d8177fa40f3de729e55d238d3c24877

nobodywasishere:
This issue should be closed due to no longer having MacOS support

PR #319, opened 2018-08-23
Made glslang as git submodule

3t0n:
Removed precompiled glslang, added submodule to fix cross build: https://github.com/ValveSoftware/Proton/issues/29 (https://github.com/ValveSoftware/Proton/issues/121), https://github.com/ValveSoftware/Proton/issues/2#issuecomment-415057704.
Updated README for MacOS and added basic manual run info.

Tested on MacOS High Sierra only, increases build time by ~30 sec.


kergoth:
As glslang is used by the dxvk build, which is done on the host, do we really need both 32 and 64 bit builds of it? We only had one before switching to a submodule, after all.

kergoth:
Off-topic, but just as an FYI, as your fork doesn't allow opening of issues :), your pre-built proton release tarball's dinput.dll.so has a hardcoded reference to the full path on your disk to libSDL-2.0.dylib, which prevents dinput.dll.so from loading into wine. It should be changed to use @rpath/libSDL.dylib or similar to align with its location next to the wine libraries. No idea how to hook that into the build process properly, though. I just fixed it manually with `install_name_tool`.

**Edit**: Looks like it should be trivial to correct the dinput issue in build_proton.sh, I'll prep a separate pull req for that.

3t0n:
Yes, you are right. 
I simple did it by template in the build script, it easily to change if someone request.
Also, it possible to reuse glslang from MoltenVK for MacOS only.

3t0n:
Thank you, I'll see what can do!

kergoth:
https://github.com/ValveSoftware/Proton/compare/proton_3.7...kergoth:osx-fixups may be of interest. Rebased a few of my local fixes on top of your branch, and merged your glslang commits, since the conversion to mac was superceded by the switch to a submodule.

kergoth:
https://github.com/wine-mirror/wine/commit/565c2dcf2aef1a555ff1025f230b4708df24d311 might be needed as well, at least my build failed without it. Did you run into that one? Of course that'd have to go in https://github.com/ValveSoftware/wine first.

3t0n:
Yes, they need to update with new wine to fix build https://github.com/ValveSoftware/wine/issues/4.

kergoth:
OT (is there a reddit or something to discuss proton on the mac? :) but could be useful: just create a  non-steam game in the mac steam client and point it to run-proton from https://gist.github.com/kergoth/5ad792c01fa7200e9eaaa7d920c4daf3 and set launch options to something like '-a 2200 quake3.exe', where 2200 is the steam appid.

It gets your list of steam libraries and searches them for the game, so you can add a library to hold the windows games as you would in linux. See the usage info for details. It can manually run executables from elsewhere with or without -a, and has an option to run proton's wine directly as a convenience to run e.g. winecfg or regedit. It also sets `STEAM_COMPAT_CLIENT_INSTALL_PATH` so proton will copy the steam dlls to connect to the mac client, and works around a dll search issue I ran into with quake 3 where it couldn't find the ones in the steam install directory.

ivyl:
Closing since support for MacOS was dropped.

PR #368, opened 2018-08-24
[MacOS] Add missing build dependency mingw-w64

tw1nk:
mingw-w64 is needed for compiling dxvk

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,10 +138,10 @@ build dxvk on your local system; refer to [the dxvk README.md](https://github.co
 Building for macOS
 ---
 To build Proton for macOS, install the latest Xcode command line tools, as
-well as cmake (for openal-soft) and a recent nasm (for libjpeg-turbo), libtool, and automake. You can
+well as cmake (for openal-soft), and mingw-w64 (for dxvk) and a recent nasm (for libjpeg-turbo), libtool, and automake. You can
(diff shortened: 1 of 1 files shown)


cjwijtmans:
this seems like a bad idea.

kergoth:
I think the better bet is just disable build_dxvk on mac, since (afaik) dxvk is useless on mac until it's able to link to MoltenVK. But this is absolutely needed to be able to build dxvk on mac, along with a few other fixes, and there's a meson bug preventing builds of dxvk 0.70 on mac right now as well.

README.md
```diff
@@ -138,10 +138,10 @@ build dxvk on your local system; refer to [the dxvk README.md](https://github.co
 Building for macOS
 ---
 To build Proton for macOS, install the latest Xcode command line tools, as
-well as cmake (for openal-soft) and a recent nasm (for libjpeg-turbo), libtool, and automake. You can
+well as cmake (for openal-soft), and mingw-w64 (for dxvk) and a recent nasm (for libjpeg-turbo), libtool, and automake. You can
```
Tetramputechture: Should be `well as cmake (for openal-soft), mingw-64 (for dxvk), and a recent nsam...`

Tetramputechture (changes requested):
If this is added, use a comma instead of an extra 'and'

aeikum:

PR #403, opened 2018-08-24
Added several dependencies for macOS to README.md

Scvairy:


flukejones:
Is mingw-w32 not also required?

josuegaleas:
The mingw-w64 package includes both.

aeikum:

PR #476, opened 2018-08-25
Update MoltenVK to v1.0.20

### LINKED ISSUE #463: Update MoltenVK to current (>=1.0.18)
Hi,
seems using MotlenVK 1.0.10 but on seems project is already at 1.0.19 on master as latest commits logs show "Update to MoltenVK version 1.0.19"..

jmannanc:
MoltenVK hasn't been updated since 1.0.10 and there have been many changes since then. You can see this doc for all of them: https://github.com/KhronosGroup/MoltenVK/blob/master/Docs/Whats_New.md

One important thing is the support of Xcode 10, amongst various fixes and improvements and getting closer to a DXVK implementation.

Fixes #463


IdanHo:
Any reason for the use of a non latest version of MoltenVK (1.0.18 instead of 1.0.19)?
It seems 1.0.19 supports Xcode 10 and comes closer to support Vulkan 1.1 which would be beneficial for Proton.

jmannanc:
An official tag of 1.0.19 hasn't been made yet, which means they might be adding more stuff to 1.0.19. If that's true I don't want to go right up to current HEAD and potentially break something that isn't supposed to be ready. 

When 1.0.19 is officially released, I can always update then too.

oscarbg:
moltenvk 1.0.20 has been released!

jmannanc:
Thank you @oscarbg, I just pushed a commit to update to v1.0.20 and I think now is the best time to merge this PR. @aeikum do you think you could take a look at this?

aeikum:
Sorry, macOS support isn't really a big priority right now and I'm super busy with other stuff so I don't have time to understand, merge, and test these changes. It might make sense for the macOS community to work in someone's fork for now so it's not bottlenecked by the official Proton review process.

jmannanc:
Is there a way for someone to test these changes and if they pass, get it merged? If not I can just go ahead and close the PR, but then I don’t understand the reason for supporting or having MoltenVK in the first place.

The fork idea could work, but if there is already support for it in native Proton, then it would be easier imo to merge MoltenVK updates.

oscarbg:
+1 on @SneakyFish5 thoughts.. if MacOS support not a big priority why don’t just merge it this PR, and if issues are found either submit to MoltenVK project and also point in README about not supported MacOS and that some breakage can occur in MoltenVK updates ( which seems unlikely).. MoltenVK updates should do more good than harm..

romulasry:
Latest release

    v1.0.21
    9ee64be

jmannanc:
@romulasry I noticed that and they really are hard at work supporting new extensions for DXVK in recent releases which is awesome. I'm not gonna update though until we get an official response on whether they'll take MoltenVK updates.

jmannanc:
MacOS support has been removed, see https://github.com/ValveSoftware/Proton/commit/a84120449d8177fa40f3de729e55d238d3c24877

Will be closing this now

PR #520, opened 2018-08-25
Make debug scripts dump dir configurable

### LINKED ISSUE #511: Proton puts executable scripts into `/tmp/proton_$USER/`
Could you make this configurable? My system does not allow to execute stuff in `/tmp` because it is mounted `noexec`.

Proton should probably put these scripts into the steam user directory by default or let me configure a tmp directory with proper permissions.

Elkasitu:
If PROTON_DUMP_DEBUG_COMMANDS is set, the directory where
the debug scripts will be dumped to can be defined with
the env variable / setting PROTON_DEBUG_DIR.

This commit also lets users set PROTON_DUMP_DEBUG_COMMANDS
in user_settings.py, which didn't work previously because
the variable was being fetched from os.environ instead of env.

Closes #511

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,7 +190,8 @@ the Wine prefix. Removing the option will revert to the previous behavior.
 | Compat config string  | Environment Variable           | Description  |
 | :-------------------- | :----------------------------- | :----------- |
 |                       | <tt>PROTON_LOG</tt>            | Convenience method for dumping a useful debug log to `$HOME/steam-$APPID.log`. For more thorough logging, use `user_settings.py`. |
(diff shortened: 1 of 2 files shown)


ryao:
It would be nice if this was documented too. It was a surprise when I spotted these scripts.

Qix-:
@Elkasitu if you wanted to go above and beyond, you could detect if the filesystem at `/tmp` was mounted as `noexec` and died with some sort of error message about setting the new env var.

Elkasitu:
I don't know if that's clearer, the user would get no apparent feedback from setting `PROTON_DUMP_DEBUG_COMMANDS` unless he also sets `PROTON_LOG` and checks the log for the error message, whereas currently the user would at least get a permission denied error.

I'll leave it up to the repo maintainers to decide, it should be a line or two of code

aeikum:
Thanks, this was merged as f9b5a0849510562489f56cb1339ce640aa659637.

PR #675, opened 2018-08-26
Minor linting fixes to build_proton.sh

Kedstar99:
--Added double quotes around various variables to avoid globbing and word splitting.
  --Removed use of legacy basetick and replaced with $()
  --Avoid use of -a and -o and replaced with || and && which are more preferred.

Suggestions and fixes were recommended by shellcheck.

The only error still suggested by shellcheck is:

Line 541:
    MAKE="make"
    ^-- SC2034: MAKE appears unused. Verify use (or export if used externally).

### DIFF
--- a/build_proton.sh
+++ b/build_proton.sh
@@ -29,7 +29,7 @@ function build_freetype
 {
     cd "$TOP"/freetype2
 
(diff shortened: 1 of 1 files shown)


aeikum:
Thanks, this was merged as b0e4ddf8e349ee85808583070e647c93bc79fd33. I removed the $STRIP changes since that variable has options embedded on macOS.

PR #793, opened 2018-08-28
[WIP] Makefile based builds

Nephyrin:
This adds an alternate build path using a `./configure.sh && make` strategy.  The configure script handles specifying the container solution to use, and the make system invokes the necessary steps within the container automatically.

Initially, this requires Docker-based steam-runtime work at ValveSoftware/steam-runtime#91, but I intend to add support for schroot containers before this is mergable.

This is a slightly more robust build strategy that allows for parallel building of projects, some more dependency tracking, and allows builds to function on non-ubuntu hosts.

### Building
Steps for producing a build with this are greatly reduced from the current build_proton.sh instructions:
- Install the mingw-w64 toolchain
- Clone ValveSoftware/steam-runtime#91
- Setup 32bit and 64bit docker images:
```bash
steam-runtime/setup_docker.sh --beta amd64 --extra-bootstrap=proton/steamrt-bootstrap.sh steam-proton-dev
steam-runtime/setup_docker.sh --beta i386 --extra-bootstrap=proton/steamrt-bootstrap.sh steam-proton-dev32
```
- Setup and build:
```bash
mkdir proton/mybuild
cd proton/mybuild
../configure.sh --steam-runtime64=docker:steam-proton-dev --steam-runtime32=docker:steam-proton-dev32
make -j12 all dist
```
- Optionally, install to steam with `make install`

### TODO
- [ ] Options for release/debug builds in configure
- [x] schroot runtime support (configure only accepts docker images at the moment)
- [x] Restore OS X build support
- [x] Add tarball package step


flukejones:
Please restore macOS build support if the intention is to replace the existing build script.

Nephyrin:
It turns out that the example commands I gave above involve running `setup_docker.sh` from outside the steam-runtime directory, which triggered a bug.  I just pushed a fix for this in the steam-runtime docker branch/pull request, so make sure you are up to date before attempting them.

Nephyrin:
I re-added the OS X specific targets and wine configuration, this should work on OS X with `../configure.sh --no-steam-runtime` but I only lightly tested things.  I also added `--osx-force-dxvk` to configure to attempt to build DXVK-via-MoltenVK, which doesn't currently work and is disabled by default.

Using docker-for-OS X it should be possible to pass a steam runtime and cross-compile for Linux/runtime hosts, but I haven't tested that, and it probably needs some tweaks.

kergoth:
FYI, minor bug in the macOS build, the freetype builds fail as they expect dist/dist/lib and dist/dist/lib64 to already exist when they try to install their libs there. Doing an mkdir -p on those before the make fixes it locally.

kergoth:
Other random notes:
- missing destdirs in the moltenvk build much like the freetype ones
- multiple cases of using a makefile rule to directly create a directory. this is bad practice, as having one rule depend on a rule whose name matches a directory path on disk results in re-running tasks anytime anything inside that directory changes. when creating directories in makefile rules, either mkdir -p them everywhere, or use a stamp file inside the directory rather than the directory itself as the rule. See https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Build_Instructions/How_Mozilla_s_build_system_works/Makefiles_-_Best_practices_and_suggestions for example, "Do not use directories as a dependency for generated targets, ever.  Any activity within a directory will alter inodes forcing targets to always be stale."
- the MOLTENVK_OUT deps for wine are being added unconditionally, they need to be added only for non-NO_DXVK, since moltenvk isn't built when dxvk isn't built
- minor note that we'll likely want to add vkd3d and build moltenvk regardless of NO_DXVK when proton updates its wine to something newer than 3.13
- I don't know if xcodebuild parallelizes by default, if not, we may want to pass `-jobs $(SUBMAKE_JOBS)` to it for the moltenvk build
- if a make fails in something after moltenvk is built, re-running make will result in moltenvk failing. its tasks seem to want to re-run when they don't need to, and the tasks aren't successfully running again
- mac builds fail without —with-ffmpeg, as wine wants avcodec, etc regardless

Claw256:
Hi, I get the following error when I enter the final line of code (make -j12 all dist)
https://pastebin.com/sJTCWe5g
I followed everything that Nephyrin said.

Nephyrin:
@Spudnik1 That sounds like you don't have the openal-soft submodule checked out.  `openal-soft/CMakeLists.txt` should exist in your proton setup.  `git submodule update --init` may help.

Claw256:
@Nephyrin Thanks for the quick reply! I'll try that now 1 sec

Claw256:
@Nephyrin Hi, I'm getting a new error after `make -j12 all dist` now
https://pastebin.com/Gq3Ju8fB
Edit: Actually it was just my bad syntax of my name of the build, it seems to have worked, but I still get this error at the end of it:
https://pastebin.com/BH6wRy7q

Nephyrin:
@Spudnik1 Looks like you passed a `--build-name` with parenthesis, which broke something.   I will look at that, but removing the parens from your configure step should fix it in the mean time.

Nephyrin:
@Spudnik1 For your second error - sounds like the error occurred further up in the build.  You'll need to look further up or run without `-j12` to see the exact problem.  It may be the bug Kergoth mentioned above, where `-j12` builds can race in creating directories (in which case building without `-j12` may just work)

Thanks for testing this ^.^

Claw256:
@Nephyrin Hi, here is another error I encountered when I ran `make all dist` (Sorry quite long)
https://pastebin.com/Tbu7GyU3

Nephyrin:
@Spudnik1 Looks like dxvk compilation failed, does your mingw64 toolchain have threading support?  I would test to see if you can build successfully with `build_proton.sh` or manually in the dxvk tree.  The dxvk repo has the specific prerequisites listed:

https://github.com/doitsujin/dxvk/blob/master/README.md

Claw256:
@Nephyrin Thanks for the info, I'm going to bed, and I'll get back to you tomorrow.

Claw256:
@Nephyrin Hi, what is the correct command to use to clone your wip-docker branch the right way?
Edit: nvm it's `git clone https://github.com/ValveSoftware/Proton.git -b makefile-buildsystem ./proton`

Nephyrin:
This has been merged into the 3.16 branch and the remaining issues brought up have (I believe) been addressed, so closing this.

Feel free to open new issues if you're encountering brokenness building on 3.16.

PR #898, opened 2018-08-29
Add a hack to determine the proper python bin to use

### LINKED ISSUE #532: Proton 3.7-4 beta - Games don't start anymore
i have switched from proton 3.7.3 to 3.7.4 beta, but all "unsupported" games don't start anymore (Gas guzzlers extreme, call of juarez, PES 2018...) 

The one that starts is Tekken 7 which was in the official list.

Python 2.7 and 3 packages are installed (verified using synaptic) as they were before but nevertheless it's probably something python related. Switching back to proton 3.7.3 games start again.

System 
Steam client version: 25/08/2018
Distribution: Ubuntu 18.04 LTS
Opted into Steam client beta?: Yes
Have you checked for system updates?: Yes
CPU: AMD Ryzen5 1500
RAM: 8GB
Video Card: NVIDIA GTX750ti
Driver version: NVidia 396.54
Video memory: 2GB

Elkasitu:
This replaces proton (python file) for a bash file that will
try to find the most appropriate python binary for executing
proton.py.

This is a workaround for some distros reserving /usr/bin/env python for
python2 only and not defining it, therefore the shebang line like above
would not work unless python2 was installed, despite the main python
proton program being written in version-agnostic code.

Fixes #532 

This works out-of-the-box, i.e. there shouldn't be any changes to be made to the way that the steam client launches proton.

Special thanks to @monzo94 as I suck at bash

P.S: Sorry for the long diff, git is stupid when renaming files


proton.py
```diff
@@ -0,0 +1,509 @@
...
+            src_file = os.path.join(src_dir, file_)
+            dst_file = os.path.join(dst_dir, file_)
+            if not os.path.exists(dst_file):
+                real_copy(src_file, dst_file)
+
+if not ("STEAM_COMPAT_DATA_PATH" in os.environ):
+    log("No compat data path?")
+    sys.exit(1)
+
+basedir = os.path.dirname(os.path.realpath(__file__))
```
Elkasitu, line 85: This is the only bit of code that I modified from the main python file, this will get `proton.py`'s directory which is what `sys.argv[0]` did previously, except this way `proton.py` can be called using `python proton.py` which wasn't possible before because the python binary is located elsewhere.

proton
```diff
@@ -1,511 +1,16 @@
...
-    log("Need a verb.")
-    sys.exit(1)
-
-sys.exit(0)
+#!/usr/bin/env bash
+# hack for distros which don't set /usr/bin/env python unless py2 is installed
+
+# get current file's dir
+dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+pybins=('python' 'python3' 'python2')
```
MayeulC, line 6: I would have favoured `python3` over `python` over `python2`.

proton
```diff
@@ -1,511 +1,16 @@
...
+
+for pybin in "${pybins[@]}"
+do
+    bin="/usr/bin/$pybin"
+    if [ -e $bin ]; then
+        echo "Proton: using $bin"
+        exec $bin "$dir/proton.py" "$@"
+        break
```
MayeulC, line 14: This isn't supposed to happen, I would print something and return an error code.

proton
```diff
@@ -1,511 +1,16 @@
...
+    fi
+done
```
MayeulC, line 16: If we reach that part, `python` isn't installed, or hasn't been found. It would be nice to tell the user about it.

proton
```diff
@@ -1,511 +1,16 @@
...
-elif sys.argv[1] == "getnativepath":
-    #windows -> linux path
-    path = subprocess.check_output([wine_path, "winepath", sys.argv[2]], env=env, stderr=lfile)
-    binary_stdout.write(path)
-else:
-    log("Need a verb.")
-    sys.exit(1)
-
-sys.exit(0)
+#!/usr/bin/env bash
```
MayeulC, line 1: I would make it a `sh` script for better compatibility.

proton
```diff
@@ -1,511 +1,16 @@
...
+# hack for distros which don't set /usr/bin/env python unless py2 is installed
+
+# get current file's dir
+dir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+pybins=('python' 'python3' 'python2')
+
+for pybin in "${pybins[@]}"
+do
+    bin="/usr/bin/$pybin"
+    if [ -e $bin ]; then
```
MayeulC, line 11: I would drop the absolute path, relying on `$PATH` instead.
We can check that `$pybin -c "print('pass')"` returns `pass` (or check the return value, but that would be less robust).

MayeulC (commented):
Looks pretty good to me,although I have a few nitpicks with the current implementation. I still think this is the better way, I'll elaborate on this a bit later.

I would have made it two commits, one with `git mv` to rename the file, then a new one with new script.

proton
Elkasitu: What isn't supposed to happen? why?

proton
Elkasitu: I went with `if [ -x "$(command -v $pybin)" ]; then` instead, which should accomplish the same task

Elkasitu:
@MayeulC I did originally use `git mv` however I fucked it up during rebase, nevertheless I redid the patch in my [branch](https://github.com/Elkasitu/Proton/tree/hack-no-python) and the PR diff still looks like a mess, but I guess that's just GitHub, the commit history for `proton.py` should be kept intact.

I can't reopen this PR since I push forced my changes, so if it ever gets considered I'll open a new one

proton
MayeulC: Sorry, I wasn't really explicit: using `exec` means that `break` won't ever be executed, unless something bad happens.

PR #919, opened 2018-08-29
Support 32-bit prefixes

### LINKED ISSUE #449: Support of 32-bit prefixes
# Feature Request

## I confirm:
      contain this feature already.

## Description
Some games only work in a 32-bit wine prefix. I couldn't find a way to create a 32-bit prefix, and creating such prefix  doesn't seem to work as well.

## References
#458 #189 #527

mat8913:
64-bit prefixes will still be created by default, this just allows Proton to use preexisting 32-bit prefixes.

Fixes #449.

### DIFF
--- a/proton.in
+++ b/proton.in
@@ -88,7 +88,7 @@ basedir = os.path.dirname(sys.argv[0])
 bindir = basedir + "/dist/bin/"
 libdir = basedir + "/dist/lib"
 lib64dir = basedir + "/dist/lib64"
-wine_path = bindir + "/wine64"
+wine_path = bindir + "/wine"
 
 #extract if needed
 dist_lock = FileLock(basedir + "/dist.lock", timeout=-1)
@@ -190,6 +190,11 @@ with prefix_lock:
         #copy default prefix into place
         mergedirs(basedir + "/dist/share/default_pfx", prefix)
 
+    if os.path.exists(prefix + "/drive_c/windows/syswow64"):
+        pfx_arch = "win64"
+    else:
+        pfx_arch = "win32"
+
     with open(version_file, "w") as f:
         f.write(CURRENT_PREFIX_VERSION + "\n")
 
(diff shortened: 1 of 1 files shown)


tingvarsson:
What are the use cases for a 32bit prefix?

MuhammedKpln:
New games are 64 bit only. Most of games are required minimum 4 gigs of ram. No need for that.

Elkasitu:
IIRC some "old" games (such as fallout 1 and fallout 2 just as an example) don't run well or at all in a 64bit prefix because the libs / frameworks required to execute them were originally created for 32bit OS and the 64bit version of these libs don't install properly on wine, or something along those lines.

tingvarsson:
Both fallout 1 and 2 works for me on 64bit. However GOG and wine-staging if that matters for those specific games.

ahallrq:
@tingvarsson .NET is another good reason. Some games use it and the workarounds for installing it into 64bit prefixes can be a bit of a pain.

Ruedii:
It would be better to just create a proper install script.

.Net now installs better on Wine64 due to Microsoft releasing the necessary 64bit/32bit universal installers for .Net. (available at their .Net developer website.)   It might be desirable to create scripts, or request from Microsoft the right to distribute a custom runtime installer.  

Microsoft is really trying to push .Net and getting .Net based games dependent on Windows-Specific .Net routines running out of box on Steam Play (and thus on the next generation of Steam Machines) will be a big plus for that.  

This is still a temporary solution.  A more permanent solution is a more complete form of wine-mono that can run almost anything that runs on .Net.

Zero86Sk:
i beleave WMP10 and WMV codecs are usable only under wine32 prefixes

elgabo86:
The Sandbox Evolution (‎466940,F2P) can't connect on network in a 64bit prefix

Arucard1983:
> @tingvarsson .NET is another good reason. Some games use it and the workarounds for installing it into 64bit prefixes can be a bit of a pain.

Specially with 32-bit games that use the 32-bit .NET framework, unless someone fix the Wine's WOW64 to make such libraries to work.

mat8913:
I've updated the pull request for proton_3.16

nsivov:
I believe fixing remaining issues with 64-bit prefixes instead is preferable.

ghost:
> I believe fixing remaining issues with 64-bit prefixes instead is preferable.

Why? There are a magnitude more problems with 64-bit prefixes than there is with maintaining code for 32-bit prefixes (judging from this PR doesn't seem like a lot.) Example: some games are currently broken because [Winetricks can't install quartz in a 64-bit prefix](https://github.com/Winetricks/winetricks/issues/1172). The fact that problems like these keep cropping up is why Winetricks suggests you try a 32-bit prefix if installing something in a 64-bit prefix doesn't work.

Tank-Missile:
Could this be updated to work with the current version of Proton? I tried every trick in the book to get a 32-bit prefix working on my own, but I had no success.

Frenzoid (approved):

PR #1087, opened 2018-08-31
proton: Choose wine binary based on executable architecture.

ghost:


Elkasitu:
In this case, wouldn't it be better to simply make the `winepath` point to `wine` by default?

Steveice10:
I wasn't sure whether there was some specific reason for using wine64 over wine, so I opted to keep 64-bit binaries using wine64.

Ruedii:
Wine64 with the WOW64 win32 comparability should work on almost all games.

There are a handful of specific games that don't function on this, and it might be good to have an option to fix it.

As a note, Wine-Mono works regardless of Win64/Win32, and .Net simply requires a different installation method.  

It is a dirty hack to simply switch to Win32 for compatibility with the Win32 .Net installation scripts.  We should instead develop scripts that work in both Win64 and Win32, and only use Win32 on games that won't work on Win64 at all.

aeikum:
Thanks for investigating this @Steveice10 , I've been wondering where that error is coming from. I'll give this a look.

pchome:
Never had any issues with simple `wine` and 32/64 bit prefixes on x86_64 system. Also, I never used `wine64`, and never noticed someone else does this.

So, maybe it's worth to simply change `wine_path = bindir + "/wine"` for now, to fix known issues?
In case this PR required more time for review, and can't be accepted ASAP.

aeikum:
This will be fixed in the next release.

aeikum:
This should be fixed by 97db8b5d8d2109a3c93e11f1d6f2e8e8957774a0.

PR #1305, opened 2018-09-07
Dynamically choose the python binary at runtime

### LINKED ISSUE #532: Proton 3.7-4 beta - Games don't start anymore

MayeulC:
This builds on my comments in #898 and #532. It should also close #532.
@aeikum, @Elkasitu and @nanonyme might be interested.

The code was tested in place of the steam-distributed proton, but only on Arch Linux, trough flatpak. The build script was not tested either.

Tell me what you think. Here are a few more comments:

# Benefits
Such an approach brings:
* broader compatibility (forward, backwards and does not require extra software)
* better error messages (helpfully tell the user that python isn't installed)
* the main executable is POSIX-compliant
* this feels less hacky than to code for platform-specific configuration, and to modify the script with sed.

# Drawbacks:
* Such a workaround is likely to stay for a very long time if merged (unless telemetry tells us it's OK to drop backwards compatibility in a few years)
* No *real* user-facing advantages over the current approach
* Needs a supplementary workaround if `proton` needs to work from the `$PATH`
* *I feel like I'm forgetting one, I'll add it back if I can remember it*

# Notes:
As-is, it prefers the binaries in the following order: `python3`, `python`, `python2`.


MayeulC:
Btw, @Elkasitu, I attributed one of the commits there to you, even if that's not *exactly* your commit, but rather a part of it. Please tell me if you are not OK with that.

nanonyme:
Maybe the one drawback you're forgetting is this is a runtime check whereas the current thing is a build-time check?

proton.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env @PYTHON_NAME@
+#!/usr/bin/env python
```
Elkasitu, line 1: is this still necessary given the proton sh script?

MayeulC:
Why would that be a drawback, though? The overhead will be negligible, depending on the needs. I timed it to 0.1 seconds in the worst case on the not-so-powerful laptop I'm typing this on.

It adds a bit of complexity, though, and one more file to the repository.

proton
```diff
@@ -0,0 +1,22 @@
...
+pythons="python3 python python2"
+
+# Get proton's path, relative to this script's
+# Note: this won't work if it was called from the $PATH
+PROTON_PATH=$(dirname "$0")/proton.py
+
+for python in $pythons
+do
+	# Try every python. The "x" is here to avoid an empty string
+	if [ $($python -c "print('success')" 2>/dev/null )x = "success"x ]
```
Elkasitu, line 13: why not something like `if [ -x "$(command -v $pybin)" ]` ? The current solution works but seems hacky

Elkasitu (approved):
LGTM, I don't think the fact that it cannot be run from path is an issue as `proton.py` assumes it is being run from the proton dir, if it ever changes to be run from path, then changing the shell script will be nothing compared to the changes to `proton.py`

proton
MayeulC: I thought about this as well, but this has the nice side effect of ensuring that `$python` really works as intended, at a marginal speed tradeoff. This doesn't really feel hacky to me, but I'm open to discussion.

proton.py
MayeulC: I think it's good practice to leave it here, and it allows directly executing the script if you feel so inclined (or if distributions end up packaging `Proton`, they can just skip the shell script and rename this). But yes, it isn't strictly necessary.

nanonyme:
I still feel this is most likely overkill. Python 2 will go end-of-life in little over a year requiring a full re-plan of this entire setup (since macOS might remove Python completely at that point) and the current thing most likely handles most if not all cases where system is reasonably new to work with Proton at all.

proton.py
```diff
@@ -84,7 +84,7 @@ def mergedirs(src, dst):
     log("No compat data path?")
     sys.exit(1)
 
-basedir = os.path.dirname(sys.argv[0])
+basedir = os.path.dirname(os.path.realpath(__file__))
```
nanonyme: This resolves symlinks. Is that intended?

MayeulC: No, that was not intended. That's a good question, and I am not sure what the correct behaviour should be, as I cherry-picked that change without giving it further thought. The only problematic case I can see is if the script itself is a symlink. If we want to retain compatibility with that case, it should be `abspath` instead. As I don't see any other problems with `abspath`, I'm going to change this. Thanks!

MayeulC:
Made mostly obsolete by a84120449d8177fa40f3de729e55d238d3c24877

PR #1327, opened 2018-09-08
Missing word

BlueBoxWare:


kisak-valve:
Hello @BlueBoxWare, this is not an error, if proton env variables are needed to get a game going, but no other workarounds are needed, then the game can still be considered for white listing.

PR #1413, opened 2018-09-12
Allow per-game user_settings

pchome:
This allow to change the runtime configuration for a specific game easier,
by copying/using `user_settings.sample.py` into
`SteamApps/compatdata/<app_id>/user_settings_pfx.py`.

I'm not familiar with python, so code is google/copy/paste.
Tried to make it 2.7/3.x compatible.

### DIFF
--- a/proton.in
+++ b/proton.in
@@ -137,15 +137,26 @@ if "PROTON_LOG" in env:
     env["WINEDEBUG"] = "+timestamp,+pid,+tid,+seh,+debugstr,+module"
 
 #load environment overrides
-if os.path.exists(basedir + "/user_settings.py"):
+if os.path.exists(os.environ["STEAM_COMPAT_DATA_PATH"] + "/user_settings_pfx.py"):
     try:
-        import user_settings
-        env.update(user_settings.user_settings)
+        sys.path.append(os.environ["STEAM_COMPAT_DATA_PATH"])
+        from user_settings_pfx import user_settings
+        env.update(user_settings)
     except e:
-        log("************************************************")
-        log("THERE IS AN ERROR IN YOUR user_settings.py FILE:")
+        log("****************************************************")
+        log("THERE IS AN ERROR IN YOUR user_settings_pfx.py FILE:")
         log("%s" % sys.exc_info()[1])
-        log("************************************************")
+        log("****************************************************")
+else:
+    if os.path.exists(basedir + "/user_settings.py"):
+        try:
+            import user_settings
+            env.update(user_settings.user_settings)
+        except e:
+            log("************************************************")
+            log("THERE IS AN ERROR IN YOUR user_settings.py FILE:")
+            log("%s" % sys.exc_info()[1])
+            log("************************************************")
 
 def check_environment(env_name, config_name):
     if not env_name in env:


aeikum:
I hesitate to add yet another place for user configuration to live and be forgotten about. You can add the variables to each game's launch options in Steam, for example `PROTON_USE_WINED3D11=1 %command%`. Is that sufficient?

pchome:
I just tried to launch a game, changing WINEDEBUG/PROTON_* variables every time it crashed.

Doing this in launch options is harder than simple edit a file.
Also I forgot to return this variables to defaults while launching another game, so decided to use such solution.

As for me the risks are the same as users now trying to use `winecfg`/`winetricks` with Proton, finally they can break the prefix but it can be safely(?) removed/recreated.

Edit: Also I don't know where I can retrieve launch options from. I had wine prefixes with Steam installed and launch options set, but now they seems lost. Files in this case are safer.

kadogo:
For me logging option never work in "game options".
So i always play with my user-setting.

I think that use a file is a good idea in case we want "backup" the options that we use.

tingvarsson:
Think we should keep tweaking simple and temporary like @aeikum suggests, and then rather look at how to permanently include known fixes per game (something like suggested in #1423).

pchome:
@tingvarsson 
This PR is not pretend to be an ultimate solution, but rather simple and quick guide for an useful feature.
It can be closed in favour of any other solution that solves described "problem".

Game's launch options is good for command line parameters, like
```
-noMovies -dx9
```
but not for long settings strings inside small text area in Steam client's dialog window
```
"WINEDEBUG="+timestamp,+pid,+tid,+seh,+debugstr,+module,+trying,+to,+figure,+out,+what,+wrong" PROTON_USE_WINED3D11=1 PROTON_NO_ESYNC=1 PROTON_WHATEVER_ELSE=1 %command% -noMovies -dx9
```

Ok, it's not what you doing regularly, but having e.g. three games (using different api) you "must" set additional parameters (PROTON_USE_WINED3D11/PROTON_NO_D3D11/DXVK_HUD/...). And using text file or maybe "rich edit" in Steam's configuration dialog window will make this process simpler.

As for #1423, it's looks more like for internal usage, database and code base will grow in time, and some mechanism to override default parameters can be handy.

tingvarsson:
@pchome I am actually always pro a config file option, easier to personally maintain IF NEEDED to tweak something. But I am hoping that these wont be needed for the options available right now, but rather be included in Proton right away so everyone gets the benefit from them.

WINEDEBUG, DXVK_LOG_LEVEL, DXVK_HUD feel temporary OR something you want in a certain way across the board.
PROTON_USE_WINED3D11, PROTON_NO_D3D11, PROTON_NO_ESYNC are debug options, and should't be considered long-term solutions. (unless they actually are, and then they should be handled directly by Proton so everyone can benefit)

In the end it is a cost vs gain discussion, not saying it is a clear case against it. (I actually thumbed up the initial proposal I remembered now..) :)

pchome:
Sure, I'm also thinking about this as an temporary solution. While Proton is Beta and many games still to be whitelisted, people experimenting with not whitelisted games, and changing game's options is a bit of pain.

This solution can be accepted for beta, and then deprecated in release.

pchome:
Closing. I'm going to try custom Proton build (as stated in readme).

pchome:
@pchome
> Edit: Also I don't know where I can retrieve launch options from. I had wine prefixes with Steam installed and launch options set, but now they seems lost.

Found in
* wine: `drive_c/Program Files (x86)/Steam/userdata/*/config/localconfig.vdf`
* native: `~/.steam/steam/userdata/*/config/localconfig.vdf`

in case someone wondering.

Issue #791, opened 2018-08-27
Train Simulator Classic (24010)


HonkingGoose:
## Train Simulator Classic

The game had another yearly update, and changed the brand name to `Classic`.

## Train Simulator 2021:

The game had another yearly update.

---

## Train Simulator 2020:

The game had another yearly update.

I no longer have access to a Linux computer to test on, so I can't provide a Proton log for TS2020.

I suspect there will still be a need to unpack the ```.ap``` files though. It seems the update to TS2020 was mainly shader work.

New hypothesis for the issue outlined in this post: having a `.` somewhere in the path causes the game to misbehave. So there might not even be a need to unpack the `.ap` files anyways...

---

### Edit 12th june 2019:

I tested a fresh install of the game, using Proton 4.2-7 and TS 2019 version 66.1b. Results:
- I still need to unpack .ap files to get all DLC to work.
- The OpenAL32.dll workaround was not needed anymore to use the 64bit version. (There used to be sound issues without workaround).
- There are still rendering issues relating to water/distant mountains.

Proton 4.2-7 log (64bit version of game): [steam-24010.log](https://github.com/ValveSoftware/Proton/files/3282536/steam-24010.log)
Environment: https://gist.github.com/HonkingGoose/7936767029a19c23d330970a1a586c45
Missing water picture: 
![Missing_water_Isle_of_Wight_DLC_at_Ryde_Pier_Head](https://user-images.githubusercontent.com/34918129/59373799-b416cc00-8d4a-11e9-884e-595beb7461d5.png)

---

The game has just been updated to the 2019 edition, which is split into 32 and 64 bit versions. So I'm updating the main post to reflect this.
I've also edited the title of the issue, because the name of the game changed on the Steam store. 
The report for the 2018 version can be seen in my edit history.

### What's new in 2019:
1. 32 bit and 64 bit versions are now available. 
1. There seems to be some kind of audio bug in the 64 bit version (for now), search for "*Setup_Audio.bat*" on the [Train Simulator 2019 FAQ](https://dovetailgames.kayako.com/Knowledgebase/Article/View/313/0/train-simulator-2019-frequently-asked-questions) for the workaround.

### What's still the same:
1. There is still a need to manually unpack **.ap** and **legacy.bin** files.
1. The same kinds of rendering/graphical glitches are still there.

#### Environment:
**OS:** Ubuntu 18.04.1
[...]

Issue #571, opened 2018-08-25
An Open Source Alternative to Corefonts


CuriousTommy:
I have been noticing that a good amount of reports here require people to install `corefonts`. While it is fine for consumers to use `winetricks` to install the `corefonts`, [the licenses](https://sourceforge.net/p/corefonts/bugs/6/#99b8) forbids Valve (or any other company) from using the proprietary fonts.

As an alternative solution, there should be a library that intercepts any request for these propritary fonts and have the program use the open source equalivent instead (credit goes to FlukyS for the idea). [I made a list of open source fonts that can be used instead on reddit](https://www.reddit.com/r/SteamPlay/comments/9a7mon/making_a_open_source_equavilent_to_corefonts/).

pchome:
More info: https://wiki.archlinux.org/index.php/Metric-compatible_fonts

ryao:
Are any of these fonts drop in replacements as far as Wine is concerned or is someone going to need to adapt them?

As for fonts, most games that require fonts need corefonts. It turns out that most of corefonts (6/11) were licensed to Microsoft by the Monotype Corporation:

https://en.m.wikipedia.org/wiki/Core_fonts_for_the_Web#List_of_fonts_and_files

That includes Arial and Times New Roman, which are the two most popular in corefonts. If it turns out that most games that need fonts need just the ones that Monotype Corporation owns, Valve might be able purchase/license the rights to them for the community. Then proton could provide the original fonts to games rather than substitutes.

By the way, if Monotype wants more money for the rights than it would cost to hire a full time developer for a year, I would rather see Valve hire a full time developer whose first task would be to fix the font situation using OSS substitutes.

CuriousTommy:
> Are any of these fonts drop in replacements as far as Wine is concerned or is someone going to need to adapt them?

The google fonts (not sure about the others) should be "metrically compatible" with the proprietary equivalent. In theory, I think it shouldn't cause issues. However, I don't think anyone has actually tested it. If I have some free time, I may try using the opensource font with Business tour (Issue https://github.com/ValveSoftware/Proton/issues/290) and see if it works.

> Valve might be able purchase/license the rights to them for the community. Then proton could provide the original fonts to games rather than substitutes.

I wonder if Valve would be willing to do that. That would definitely make everyone's life easier.

With that being said, I still would also like to have an open source solution so that it is easier to distribute the font to any wine-based install without restrictions.

[...]

PR #1423, opened 2018-09-13
Implement mechanism for game-specific tweaks and workarounds

dreamer:
Quote from tweak module docstring:

> This module lists game-specific tweaks, that cannot be handled by other means (specifically tweaks, that are outside of dxvk.conf or Mesa scope, such as additional environment variables, renaming files post-installation, changing Wine settings for the specific game, etc.).  The purpose is similar to "Install Scripts" *.vfd files distributed with games, except here tweaks are provided by Proton community and not game developer.

To show usage, I included workarounds for three games:

* STAR WARS™ Jedi Knight - Jedi Academy™, appid 6020
* STAR WARS™ Jedi Knight II - Jedi Outcast™, appid 6030
* Call of Duty® (2003), appid 2620

After these fixes, Jedi Academy could be whitelisted, Jedi Outcast requires additional work (I have no sound and certain id Tech 3 - specific workaround needs to be injected to fix "grey filter" bug). I don't own Call of Duty (I included fix based on [reports on SPCR](https://spcr.netlify.com/app/2620)).

I tested Star Wars titles on two systems: [NVIDIA](https://gist.github.com/dreamer/03179db6f82b01e06ec8fa41faf09f00) and [Mesa/Intel](https://gist.github.com/dreamer/4f3773020ded88d4612e41de4dd67b7e).

pylint reports 10/10 on tweaks.py module

I have additional functionality for Tweaks class in development (importing Core Fonts installed in users' OS, which would help to run multiple titles, including e.g. GTA V). I decided to not include more functionality with this PR, because it needs more testing and I want to learn if Proton maintainers are interested in this mechanism at all.

Some ideas, how this class could be expanded in future:

* switching Wine windows version in appid prefix
* injecting new default settings into games
* providing game-specific defaults for Proton environment variables
* installing additional dependencies missing from Install Script included in windows depot


dreamer:
Oh, maybe I should include a short list of titles, that *could* run better if started with some additional environment variables (compiled based on SPCR reports):

* Battle Brothers
* Battlerite
* Bioshock
* Bioshock 2
* Claire
* LEGO® Star Wars™ - The Complete Saga
* LEGO® Indiana Jones™: The Original Adventures
* Sonic Generations Collection
* Thief (2014)

... and probably many more. When browsing through reports, the most common workarounds I saw included: disabling esync, other OpenGL related flags and PulseAudio related flags.

If Tweaks included functionality installing Core Fonts in game prefix, it would significantly improve the situation (there are many situations where games do not respect font replacements provided by wine and in result either game is unplayable or various game launchers cannot start). Also, Core Fonts installed in prefix solve issue, where Install Script tries to install vcredist libraries and keeps failing.

ihusky02:
Did you can make this tweak for world of tanks blitz? (Ap ID 444200)

nanonyme:
Am I misunderstanding or is this hardcoding an nVidia proprietary driver hack most likely done because they implemented core profiles wrong into Proton?

dreamer:
> Did you can make this tweak for world of tanks blitz? (Ap ID 444200)

This game worked for me perfectly without any tweaks. What exactly do you have in mind?

nanonyme:
Ah, there's also Mesa-specific hacks. https://www.mesa3d.org/envvars.html you might want to document in the change that this is a workaround to too small fixed size buffer for reading gl extensions in the game
Same is most likely the case with the nVidia-specific value

dreamer:
> Am I misunderstanding or is this hardcoding an nVidia proprietary driver hack most likely done because they implemented core profiles wrong into Proton?

No, that's not it. There is a bug in the game engine, which was fixed sometime after these games were released.

Place of OpenGL-related tweaks like these would be in OpenGL implementation, so outside of Proton but rather in Mesa or NVIDIA driver. While we theoretically could implement a workaround in Mesa, it would affect only AMD users - it wouldn't work for Intel users, because their drivers do not have per-binary workarounds, and it wouldn't work for NVIDIA users, because if NVIDIA didn't include it so far, they will never do it in their Linux driver.

The way I am proposing in this PR, it will work for everyone using Proton, out of the box.

> (...) you might want to document in the change that this is a workaround to too small fixed size buffer for reading gl extensions in the game. Same is most likely the case with the nVidia-specific value

I already did document it, in [commit message](https://github.com/ValveSoftware/Proton/pull/1423/commits/23f6a28a8b04dc1a4857d93ad2dbbb6770a07852) implementing the change :)

ihusky02:
> > Did you can make this tweak for world of tanks blitz? (Ap ID 444200)
> 
> This game worked for me perfectly without any tweaks. What exactly do you have in mind?

Game shop, tournaments, messages section don't work ( when click on this, game sudenlly close). I think this hapen because net framework and corefonts are not added in proton or in game prefix.

dreamer:
> > > Did you can make this tweak for world of tanks blitz? (Ap ID 444200)
> > 
> > This game worked for me perfectly without any tweaks. What exactly do you have in mind?
> 
> Game shop, tournaments, messages section don't work ( when click on this, game sudenlly close). I think this hapen because net framework and corefonts are not added in proton or in game prefix.

Then I suggest installing corefonts using [protontricks](https://github.com/Sirmentio/protontricks) to test if it'll resolve the issue for you: `$ protontricks 444200 corefonts`.

I'll be monitoring #1293 to see if I can include fixes for this particular game in future patches.

Let's keep discussion here focused on general mechanism and not specific games :)

nanonyme:
Sorry, missed that somehow. The mechanism does indeed sound sensible

ihusky02:
> > > > Did you can make this tweak for world of tanks blitz? (Ap ID 444200)
> > > 
> > > 
> > > This game worked for me perfectly without any tweaks. What exactly do you have in mind?
> > 
> > 
> > Game shop, tournaments, messages section don't work ( when click on this, game sudenlly close). I think this hapen because net framework and corefonts are not added in proton or in game prefix.
> 
> Then I suggest installing corefonts using [protontricks](https://github.com/Sirmentio/protontricks) to test if it'll resolve the issue for you: `$ protontricks 444200 corefonts`.
> 
> I'll be monitoring #1293 to see if I can include fixes for this particular game in future patches.
> 
> Let's keep discussion here focused on general mechanism and not specific games :)

Tell me how to do this? I need only write this command in terminal?

dreamer:
@ihusky02 I left a comment for you in #1293 

@kisak-valve  @aeikum @Nephyrin What do you think about this PR? Should I keep working on this or do you prefer some other solution?

tweaks.py
```diff
@@ -0,0 +1,71 @@
...
+    "6030": {
+        "env": {
+            "MESA_EXTENSION_MAX_YEAR": "2003",
+            "__GL_ExtensionStringVersion": "17700",
+        }
+    },
+}
+
+
+class Tweaks:  # pylint: disable=too-few-public-methods
```
Elkasitu: Unless I'm missing something, this is unnecessary, you can just export `TWEAKS_DB` and do `env.update(TWEAKS_DB.get(env.get("SteamAppId")).get("env", {}))`

The pylint error about too few public methods is a hint :wink:

tweaks.py
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env python
+
+# This is free and unencumbered software released into the public domain.
```
Elkasitu: I doubt the compatibility of this license with Proton's

Elkasitu (commented):
This is a good idea, but perhaps a better implementation would be to fetch the tweaks from a real online db, and users could activate tweaks from the steam client ui (something like "automatically improve game compatibility") but I know that this requires infrastructure, Valve (or someone with enough resources) would need to set up a "wine game tweak db" that can be queried once during game setup/update, in this manner new game tweaks could be used **without updating Proton**, but this is a good start.

Maybe we could use a google document / github repo as a nosql db, or something like that

Just a quick suggestion: Tweaks can be defined in a data format instead of a source code format (python) such as `csv`, `json` or even `xml`, there's no logic in the `.py` file and the aforementioned formats have the advantage of having GUIs for easily performing CRUD operations, this is not the case for python.

tweaks.py
dreamer: Oh yes, I know - as I explained in PR, I am developing additional functionalities to this class, and this warning is already disabled in my local checkout :)

tweaks.py
dreamer: I copied it from `filelock.py` distributed with Proton. But I probably should change hashbang to require `python3`.

tweaks.py
nanonyme: This looks reasonably hard to read in such a packed form though

dreamer: @nanonyme I agree. Putting this line straight into `proton` file would also make it harder to implement unit tests in future. And Proton definitely could use some tests right now...

Elkasitu:
```python
game_tweaks = Tweaks(env.get("SteamAppId"))
env.update(game_tweaks.env)
```
to

```python
game_tweaks = TWEAKS_DB.get(env.get("SteamAppId"), {})
env.update(game_tweaks.get("env", {}))
```

is as readable, you could use even more intermediate variables to make it super readable, if you wanted to

dreamer:
@Elkasitu 
> would need to set up a "wine game tweak db" that can be queried once during game setup/update, in this manner new game tweaks could be used without updating Proton

I think it might be overkill - dxvk tweaks are already distributed with Proton updates and it might be beneficial to keep this file in sync with wine and dxvk versions (maybe a new version of Wine will turn some workarounds moot?). By keeping this in repo, we create an easy way for the community to engage implementation of tweaks (e.g. PRs can be referenced from Whitelist requests).

Also, tweaks might need to be implemented not only during setup/update but maybe on every start of the game (it's already the case for using Core Fonts installed in OS).

> Maybe we could use a google document / github repo as a nosql db, or something like that

Anything but Google Docs, please :) I was helping with [SPCR](https://spcr.netlify.com/) Google sheet and it, erhm, "didn't work very well" (SPCR almost migrated away from gsheets at this point).

> Just a quick suggestion: Tweaks can be defined in a data format instead of a source code format (python) such as csv, json or even xml, there's no logic in the .py file and the aforementioned formats have the advantage of having GUIs for easily performing CRUD operations, this is not the case for python.

That was my idea (and first implementation) as well, but decided to switch to simple python dict for the following reasons:

* it's simpler, less code, fewer opportunities for error
* it's faster (but I can't imagine this database to grow too big, so it's very tiny advantage)
* in dict we can include python functions as values - this might prove useful
* Only tweaks, that are required for a game to run/not_crash/provide_full_functionality should be included here. Performance tuning and such should be left for users to make on their own. Games should just work, without setting up checkboxes in Steam UI or using a separate tool like `protontricks`. I imagine sometimes a tweak will improve a game for some users and make it worse for others - in such cases, it should be excluded from this file.

tweaks.py
dreamer: ... unless I want to use `game_tweaks.import_fonts()` in some other place in code ;) But I hear you - maybe it would be better to implement future tweaks as more functions in `proton` file, but I wanted to separate responsibilities early. `proton` script is long and non-pythonic in several ways already.

ihusky02:
protontricks don`t work so good. I can`t install vcrun 2012 and 2015 because  it`s don`t change wineprefix to 32 bit (for Ap ID 444200) and I have returned error status 5 with vredist_x86.exe ehhhh  I don`t know why it work when I use Wine and install steam client use winetricks but don`t work with native steam for linux and protontricks.

Ahhh and add information that protontricks work when winetricks is copied into /usr/bin/ directory and STEAM_DIR AND PROTONVERSION is set

nanonyme:
I do hope it's enough that it's in PATH. If not, can you link to source so I can do PR?

ihusky02:
> I do hope it's enough that it's in PATH. If not, can you link to source so I can do PR?

No I think it`s because steam play or protontricks don`t do something like winetricks (WINEARCH=win32 WINEPREFIX=/home/_youruserdir/.wine winecfg) <=== create 32bit prefix before install important things and game. Otherwise I don`t know where is place steam play wine, because I can`t get into path /Proton 3.7/dist/ because I have error "too many arguments" or file or directiory doesn`t exist.

dreamer:
@ihusky02 this thread is for discussing Pull Request, not usage of `winetricks` or `protontricks`. Continue this discussion in #1293, please.

dreamer:
I updated Tweaks with the ability to import corefonts from OS. Quote from commit message:

> When trying to run a game, that depends on corefonts to run correctly, Proton looks for corefonts installed in OS and hard link or copy them to expected location in a game prefix.
>
> No additional registry entries are required (at least for titles, that I tested so far), Wine automatically registers these fonts when a game starts.
>
> Tweaks module tries to find fonts using fc-list (for system and user locations), then it looks into default Wine prefix (~/.wine/), and finally into /usr/share/fonts/* locations, where corefonts are likely to reside.

I scanned through [SPCR reports](https://spcr.netlify.com/raw-reports) and included tweaks for games, that I am quite sure will benefit from this change (tested some of those titles to confirm).

Next step would be to bundle Proton with FLOSS replacements for Windows fonts (the issue is discussed here: #571), at packaging time install them in `dist/share/default_pfx` and register as replacements for their original counterparts (this is especially important for non-western fonts). After that, we could trim `TWEAKS_DB` to include only titles, that depend on original fonts.

@kisak-valve @aeikum I am waiting for your response about this PR - do you like where this is going?

tweaks.py
dreamer: Changed to python3 (I am not going to write code that works in both python2 and python3 at the same time, it's a hassle).

ihusky02:
> I updated Tweaks with the ability to import corefonts from OS. Quote from commit message:
> 
> > When trying to run a game, that depends on corefonts to run correctly, Proton looks for corefonts installed in OS and hard link or copy them to expected location in a game prefix.
> > No additional registry entries are required (at least for titles, that I tested so far), Wine automatically registers these fonts when a game starts.
> > Tweaks module tries to find fonts using fc-list (for system and user locations), then it looks into default Wine prefix (~/.wine/), and finally into /usr/share/fonts/* locations, where corefonts are likely to reside.
> 
> I scanned through [SPCR reports](https://spcr.netlify.com/raw-reports) and included tweaks for games, that I am quite sure will benefit from this change (tested some of those titles to confirm).
> 
> Next step would be to bundle Proton with FLOSS replacements for Windows fonts (the issue is discussed here: #571), at packaging time install them in `dist/share/default_pfx` and register as replacements for their original counterparts (this is especially important for non-western fonts). After that, we could trim `TWEAKS_DB` to include only titles, that depend on original fonts.
> 
> @kisak-valve @aeikum I am waiting for your response about this PR - do you like where this is going?

can you tell me how use this script?

dreamer:
> can you tell me how use this script?

The proper way would be to [build Proton](https://github.com/ValveSoftware/Proton#building-for-linux) from [my branch](https://github.com/dreamer/Proton/tree/po/tweaks) and replace the version downloaded automatically by Steam (that's the way I do it). But I tested my changes with WoT:Blitz, it did not resolve crashes in that game (that's why it is not included in `TWEAKS_DB`).

ihusky02:
> > can you tell me how use this script?
> 
> The proper way would be to [build Proton](https://github.com/ValveSoftware/Proton#building-for-linux) from [my branch](https://github.com/dreamer/Proton/tree/po/tweaks) and replace the version downloaded automatically by Steam (that's the way I do it). But I tested my changes with WoT:Blitz, it did not resolve crashes in that game (that's why it is not included in `TWEAKS_DB`).

Talk me more, when I am green user like total noob, so I must do this command on clean install native steam  client (without enable steam play)? But how to replace proton version?

OK, let me know if you find solution my friend, good luck :) and have a nice day

dreamer:
Implemented additional tweaks feature: commandline parameters for specific binary. Also, introduced `PROTON_NO_TWEAKS` variable, so user can override all tweaks in one place and touched few lines of code in `proton.in` to make it more... python ;)

@aeikum  still waiting for *any* response regarding this PR

ihusky02:
World of Tanks Blitz still doesn`t work with this tweaks?

@dreamer check message from stanislawtlolka@gmail.com

aeikum:
@dreamer It's a complicated question. We already have a mechanism for this kind of thing (the steam compat config variable), so we need to evaluate the pros and cons of doing it this way instead. I don't have an answer for you yet, sorry.

dreamer:
@aeikum The only problem I have with `STEAM_COMPAT_CONFIG` is lack of public access to the database - I can't browse it to see what fixes work for what titles, and (more importantly), I can't suggest fixes through PR or other processes (I guess Whitelist issue template exists for that purpose, maybe? But it covers only already existing `PROTON_*` variables).

I understand you don't want to complicate things and duplicate existing functionality - so I would be fine with dropping `TWEAKS_DB` variable and exposing specific tweaks through additional `PROTON_*` variables.

Another way forward could be for me to rename `PROTON_NO_TWEAKS` to `PROTON_COMMUNITY_TWEAKS` (disabled by default) - this way `tweaks.py` could serve as staging place for community PRs, before you adopt fixes into "steam compat config" mechanism.

tweaks.py
```diff
@@ -60,12 +65,110 @@
...
+            ('Verdana', 'Bold Italic', 'verdanaz.ttf'),
+            ('Webdings', '', 'webdings.ttf')
+        ]
+        for font in ms_core_fonts:
+            if not self.is_font_installed(font[0]):
+                self._install_ttf_font(*font)
+
+
+    def is_font_installed(self, file_name):  # pylint: disable=missing-docstring
+        return os.access(self.prefix + '/drive_c/windows/Fonts/' + file_name, os.F_OK)
```
plaes: os.path.join

dreamer: I know `os.path.join` is an idiomatic way of creating paths in Python, but:

1) Rest of code uses string concatenation for this purpose
2) This code is not cross-platform, it is intended to run on Unix systems only
3) There are no user-supplied strings concatenated here - we can expect prefix and file_name to be properly escaped already (this is a shaky argument)

*I am not ruling out using os.path.join, merely explaining why I didn't use it for this PR.*

ihusky02:
@dreamer something new, about run word of tanks blitz?

flibitijibibo:
Another possible use for this is the use of XnaToFna to fix games that depend on XNA to run; wine-mono doesn't cover C++/CLI assemblies so at the moment it's impossible to run these games without the .NET runtime. The way we have it set up, you target an exe file and it relinks the dependencies to use FNA instead, while also covering a couple formats we can't support (plain WMV/WMA files), and this removes the C++/CLI issue for most XNA games out there.

This is something I'm [looking into now](https://github.com/0x0ade/XnaToFna/issues/13), but the idea is that a really small zipfile can run XnaToFna on games known to depend on XNA (unfortunately the redists can be added as deps even if this isn't the case, so it'd be a manual process), then the game can run through Wine as usual.

Mind you, this doesn't cover every XNA game - only 4.0 is supported, and games using C++/CLI in any other capacity are out as well (old Steamworks C++/CLI wrappers, DirectInput wrappers, etc). The long-term solution would be to implement C++/CLI but that's a multi-year project for an _extremely_ small percentage of games out there.

This is a comically extreme example of how a game-specific tweak tool could be used, but it's worth a thought.

lucifertdark:
The game [586970] Hungry Flame needs Corefonts as well, other than that it works perfectly. I've added a report to SPCR just now but how long it'll take to show up, who knows?

tweaks.py
nanonyme: I'd say the more interesting usage of os.path.join is avoiding double-path-separators. Forward slash works just fine on Windows even

dreamer:
Summary of changes:

* Rebase patches on top of current `proton_3.7` branch.
* Use rebase as an opportunity for some stylistic changes to tweaks.py (using `'` for strings consequently)
* Use os.path.join throughout tweaks.py
* Add functionality for creating .ini files in game directory and used it for creating override for dosbox configuration file for Tomb Raider I (224960)
* Add Hungry Flame (586970) to list of games needing Core Fonts installation

tweaks.py
dreamer: @plaes @nanonyme - implemented, thank you for review :)

dreamer:
@flibitijibibo I tried to make Duck Game work with XnaToFna to possibly integrate it in tweaks, but had no success so far - I suspect a problem with my mono installation.

@kisak-valve @aeikum please take a look at dosbox fix specific for Tomb Raider in commit [ba97fe](https://github.com/ValveSoftware/Proton/pull/1423/commits/ba97feab99d17864201fdc0a95c09ee9db251e1a) - I don't think these kind of fixes are feasible with steam compat config mechanism. Did you have a chance to discuss internally if this way of distributing workarounds for games fits Proton?

HonkingGoose:
@dreamer you said to drop a comment if anybody finds a game with font issues? I have a game which has font issues for some people:

Could someone who understands fonts take a look at [this](https://github.com/ValveSoftware/Proton/issues/791#issuecomment-416818307)? A user has font related problems with Train Simulator (2018 version at the time of report). I don't have font related issues, but I use a standard Ubuntu 18.04 install.

I'm not entirely sure if this is relevant, but I see a lot of mentions with regards to fonts around here, so this seemed a good place to ask. If not feel free to tell me its not relevant here, and I'll delete the comment to keep things clear on this issue.

dreamer:
@HonkingGoose Existing corefonts workaround does not help in Train Simulator, so I can't include it here. I'll keep tabs on this problem, however - let's continue the discussion in #791.

====

I am closing this PR in favour #1745 (updating this one with rebased commits would create too much mess). Also, I believe Proton devs would prefer to stabilize 3.7 branch for games, that were already whitelisted so this PR is unlikely to be accepted.

PR #1488, opened 2018-09-16
Update PREREQS.md

bendavisme:
Added one more step to setting up nvidia drivers. This may just be specific to Ubuntu.


kisak-valve:
[nvidia-driver-###](https://packages.ubuntu.com/bionic/nvidia-driver-390) is a meta-package which should pull this as a dependency. If it doesn't, then this is a packaging bug that should be reported to the PPA or distro package maintainer.

Issue #1510, opened 2018-09-19
How to build on debian?


bstin:
Sorry for the basic nature of this question, but I've followed the runtime / proton guides and have my chroots built.  However, I can't actually even start the build process?

My root dir is as follows:

``````
root@debian9:~# ls
proton	
steam-runtime
``````

Entering `cd proton; ./build_proton.sh` results in:

```
./build_proton.sh: line 562: [: : integer expression expected
./build_proton.sh: line 562: [: : integer expression expected
E: Failed to change to directory ‘/root/proton/build/openal.win32’: No such file or directory
I: The directory does not exist inside the chroot.  Use the --directory option to run the command in a different directory.
```

It appears that line number (562) is trying to get the version number from gcc for amd64/i386 chroots. I've verified if I `schroot --chroot steamrt_scout_beta_amd64` I can then run gcc and it is the correct gcc-5 version.

I'm unsure where to go from here?

dreamer:
I had the same issue - in my case schroots environments had no access to directory with Proton source code. I solved it by reconfiguring `/etc/schroot/default/fstab` to bind source code directory.

In my case it was not issue with Proton, though - I build Proton in Fedora (through Debian running in systemd container) and my source code was already bound once by systemd (so schroot inside container did not see dir at all). If you think it might help you, I can publish howto on wiki or somewhere...

aeikum:
Hm, what version of Debian are you running?  What is the output of this command when run as your user? `schroot --chroot steamrt_scout_beta_amd64 -- gcc -v` Also, you don't want to build Proton (or anything) as root...

bstin:
@dreamer / @aeikum 

Thanks to both of you. 

For anyone arriving here by google, there are two solutions:

1. Get your `/etc/schroot/default/fstab` sorted out per @dreamer recommendations 
2. Easier solution (as @aeikum suggest) , is don't build as root. Build as normal user since all of schroot seems to expect this. Make sure you've done following:

- sudo is installed
- you've run `visudo` and added your user permissions
- added your user to `sudo` group and can intelligently run commands via sudo (ie. `sudo ls -l` works)

PR #1530, opened 2018-09-21
Prevent users from running build_proton.sh as root

dreamer:
Prevents users from borking their systems and wasting time wondering why schroot does not work, as per #1510

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 dist/
 build/
 contrib/
+*.swp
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1,3 +1,4 @@
 dist/
 build/
 contrib/
+*.swp
```
kisak-valve, line 4: Unrelated?

dreamer: I didn't think it merits its own, separate pull request.

aeikum:
I dunno, not doing things as root is pretty much Unix 101 and this might prevent someone who does know what they're doing from doing something they want to do.

PR #1552, opened 2018-09-23
Clearify Proton is based on the Linux Client, remove macOS

ShalokShalom:
macOS is officially not supported: 
https://steamcommunity.com/games/221410/announcements/detail/1696055855739350561

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
 ---
 Introduction
 ---
-Proton is a tool for use with the Steam client which allows games which are
-exclusive to Windows to run on Linux and macOS operating systems. It uses Wine
+Proton is a tool for use with the Linux Steam client which allows games which are normally 
+exclusive to Windows to run on Linux as well. It uses Wine
 to facilitate this.
 
 Most users will prefer to use Proton provided by the Steam client itself.  The


Jacalz:
I believe that Proton itself is supported on Linux and MacOS but the implementation in the Steam client is only intended for use on Linux. Since it’s based on Wine it still works to build and use on MacOS too...

ShalokShalom:
So Proton works without a Steam Client? Or is it supported with the macOS client, while not being implemented?

Jacalz:
I believe it does, might be wrong, but I am fairly certain. I am not affiliated with Valve in any way, but I think that it goes as far as just not being in the steam client due to reasons inside Valve. Might be that Proton is only meant for making Windows games playable when Microsoft pulls the plug on Win32 apps. Just speculation from my side, but if that's the case, then it would mean that moving away from MacOS would be a smart decision too. Especially due to Apple removing OpenGL in Mojave...

Gcenx:
> So Proton works without a Steam Client? Or is it supported with the macOS client, while not being implemented?

Proton does work without the Steam Client

The issue you will have is if you want to use a Steam game you would need to revert the Isteamclient hacks so your able to launch a Windows Steam Client. 
I already did the above in my person fork of the wine portion.


> I believe it does, might be wrong, but I am fairly certain. I am not affiliated with Valve in any way, but I think that it goes as far as just not being in the steam client due to reasons inside Valve. Might be that Proton is only meant for making Windows games playable when Microsoft pulls the plug on Win32 apps. Just speculation from my side, but if that's the case, then it would mean that moving away from MacOS would be a smart decision too. Especially due to Apple removing OpenGL in Mojave...

Apple is officially depreciating OpenGL on Mojave it was not removed. Try it yourself Mojave still runs normal Wine, it will just complain that you ran a 32bit application.

If Valve didn’t plan to support macOS eventually they’d would have removed all mentions from here & not had MoltenVK in the project as Linux does not require it, plus CodeWeavers has an employee focused on gettin VKD3D working via MoltenVK so it’s only a matter of time until it’s viable to run on macOS. (32bit removal is also being worked on)

jmannanc:
Judging by this comment, Valve doesn't have a firm stance on macOS support: https://github.com/ValveSoftware/Proton/pull/476#issuecomment-418356496

ShalokShalom:
I recommend clearing this descrption in the Readme up, anyway how. Currently, its quite confusing to someone who is not involved into the project. Since there is no "Steam Client" and because Proton is not running on the Windows version, I guess its sane to add "Linux" here. And when the macOS client is essencially supported and has no maintainer currently, also add that one.

flukejones:
I am against this change as Proton compiles and runs on macOS just fine, and is actively worked on.

jmannanc:
Actively worked on may not be true.

Gcenx:
> I am against this change as Proton compiles and runs on macOS just fine, and is actively worked on.

Works just fine is not exactly true ISteamClient doesn’t map the SteamAPI calls to macOS calls only Linux currently. The usefulness of this is limited without reverting the related ISteamClient hacks within there wine fork.

If we at least had working ISteamClient mapping I wouldn’t care that currently Proton is only really on working directly from Steam using the Linux client.

ShalokShalom:
Currently, its simply confusing.

ryao:
I did not find it confusing.

jogolden:
not confusing

flukejones:
Proton is separate to the Steam client.

If anything I'd change the wording to:

> Proton is a tool which allows games which are normally exclusive to Windows to run on Linux and macOS operating systems. It uses Wine to facilitate this. Proton can be run officially via the Linux Steam client, and separately on macOS with official macOS Steam client support to come later.

or something similar.

ryao:
Mac OS X support has been removed:

https://github.com/ValveSoftware/Proton/commit/a84120449d8177fa40f3de729e55d238d3c24877

Jacalz:
Yeah, as of 3.16 MacOS support is removed and I think that this can be closed then...

flukejones:
Looks like that is part of a larger series of changes which hopefully should include macOS again - https://github.com/ValveSoftware/Proton/pull/793

cjwijtmans:
clarify*

aeikum:

PR #1619, opened 2018-09-27
Add support for dxvk d3d10 dlls to proton

tingvarsson:
Note: Not tested yet as I am just about to setup the additional things needed to build a proton package.

But wanted to get input on why this hasn't been added yet? Suspecting that the d3d10 support isnt fully trusted yet or something? Please comment if I'm wasting my time testing this. @aeikum @doitsujin

### DIFF
--- a/build_proton.sh
+++ b/build_proton.sh
@@ -469,11 +469,17 @@ function build_dxvk
     mkdir -p "$DST_DIR"/lib64/wine/dxvk
     cp "$TOP/build/dxvk.win64/bin/dxgi.dll" "$DST_DIR"/lib64/wine/dxvk/
     cp "$TOP/build/dxvk.win64/bin/d3d11.dll" "$DST_DIR"/lib64/wine/dxvk/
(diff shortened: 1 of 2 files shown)


aeikum:
> But wanted to get input on why this hasn't been added yet?

A change like this affects many, many applications and needs to be thought about and QAed. We have been focusing on other things. We'll be approaching DXVK in another way in the near future (building as native Linux libraries instead of Windows DLLs), so we won't be using the approach you've used here, but I don't see anything wrong with it if you'd like to use it in your builds in the meantime.

tingvarsson:
Havent seen or looked for numbers but cant be that many as dx10 wasnt a huge success. And i was hoping that dxvk would be the goto solution for those games as well. :) but realize that your approach to fallback to dx9 is most likey the wiser in most cases as it should be more stable.
And it is probably only a few titles that would need the benefit from the performance gain. (E.g. crysis, just cause 2).

PR #1653, opened 2018-09-30
Update PREREQS.md

ZMASLO:
DXVK and proton wasn't working on my Radeon R9 290 and I figured out why. I provided solution for problem (many people had it) and also I provided verification that Vulkan is working by running command vulkaninfo.

I hope that I helped.

### DIFF
--- a/PREREQS.md
+++ b/PREREQS.md
@@ -28,6 +28,22 @@ sudo apt install mesa-vulkan-drivers mesa-vulkan-drivers:i386
 
 Provide your user password when requested and reboot after the last command completes to ensure the driver has updated correctly.
 
+On Radeon R9 200/300 series you have to blacklist radeon module and add following parameters to amdgpu:
+
+```
+echo "blacklist radeon" | sudo tee --append /etc/modprobe.d/blacklist.conf
+echo "options amdgpu si_support=1 cik_support=1" | sudo tee --append /etc/modprobe.d/amdgpu.conf
+sudo update-initramfs -u
+sudo reboot
+```
+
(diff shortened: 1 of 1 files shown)


PREREQS.md
```diff
@@ -28,6 +28,22 @@ sudo apt install mesa-vulkan-drivers mesa-vulkan-drivers:i386
 
 Provide your user password when requested and reboot after the last command completes to ensure the driver has updated correctly.
 
+On Radeon R9 200/300 series you have to blacklist radeon module and add following parameters to amdgpu:
+
+```
+echo "blacklist radeon" | sudo tee --append /etc/modprobe.d/blacklist.conf
```
nanonyme, line 34: Might want to have a warning here. It's not enabled by default because it's not done even with latest kernel releases

kisak-valve:
Hello @ZMASLO, this has moved to https://github.com/ValveSoftware/Proton/wiki/Requirements.

codeman101:
Why are the instructions in the PREREQS.md only for ubuntu users? What about Fedora, Arch and OpenSuse?

omicron-b:
@codeman101 from https://store.steampowered.com/about/ 
```
Supported Operating Systems

Windows 7, 8, 8.1, 10, or later.
Intel Mac, OS X version 10.11 (El Capitan), or later.
Ubuntu Linux (latest version).
```

I am not saying that the instructions should not be added by the community, but that is probably why they were not added by Valve.

aeikum:
This stuff should go into the wiki now.

PR #1698, opened 2018-10-05
[RFC] Possible wined3d11 symlink problem

kakra:
While testing, I found that Proton created symlinks to non-existing files in the wined3d path. Shouldn't the files be linked to fakedlls instead? I don't see any `*.dll` directly in `lib/wine` here.


kakra:
I think this has been fixed in the beta. Closing.

Issue #1711, opened 2018-10-09
Gorogoa (557600) minor audio crackle with music


dreamer:
# Compatibility Report
- Name of the game with compatibility issues: Gorogoa
- Steam AppID of the game: 557600

## System Information
- GPU: Intel Sandybridge Mobile
- Driver/LLVM version: 3.0 Mesa 18.0.5
- Kernel version: 4.18.11
- Link to full system information report as [Gist](https://gist.github.com/): [1](https://gist.github.com/dreamer/4f3773020ded88d4612e41de4dd67b7e)
- Proton version: 3.7-7 Beta

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

## Symptoms 

Note: my GPU does not support Vulkan, therefore I needed to start game with: `PROTON_USE_WINED3D11=1 %command%`. On my desktop system (with NVIDIA GTX770), game works out of the box.

Overall this game is very close to "whitelistable" state, but *sometimes* refuses to start (it's hard to reproduce, though). Also, slight sound crackling is present.

## Reproduction

Just install and start the game. I can *sometimes* (very rarely) reproduce game hanging (eventually Gnome offers me to close the unresponsive window in such cases). Next time I'll manage to reproduce, I'll upload a log. Sound crackling happens almost always, but it's not a big deal.

kisak-valve:
Hello @dreamer, this game was found in the 2019-07-31 whitelist update. Are you still experiencing this issue with Proton 4.2 or newer?

dreamer:
Just tested with 4.11-1. On my old laptop, I still need `PROTON_USE_WINED3D11=1` to start the game, so I guess you didn't define it on your end in compat database. On Vulkan-capable hardware it worked fine out of the box. Game did not hang on start even after several attempts - so this part seems improved.

I experienced no change regarding music - tiny sound crackling is still there (happens mostly when music occasionaly is being loud - which is rare in this game). This does not affect gameplay in any way and is not really very distracting. If you want to pursue this issue, I can upload details of my sound configuration (but it is basically default setup for Fedora, I didn't mess with PulseAudio config files).

As for the gameplay - some time ago I finished this game successfully (using Proton 3.16-4 based on my report on protondb) and didn't experience any game-breaking bugs. There are more reports of people finishing it using Proton 4.2-x.

[...]

Issue #1094, opened 2018-08-31
[whitelist] Virginia (374030)


ziabice:
Tried with Proton 3.7-3.

The game worked flawlessy. 

My specs: [https://gist.github.com/ziabice/8c19008e58f82b16a62f7ab65ba235d4](https://gist.github.com/ziabice/8c19008e58f82b16a62f7ab65ba235d4)

KGOrphanides:
No joy here. Logs attached - looks like a graphics issue with my once-powerful but decidedly ancient graphics card.

[error.log](https://github.com/ValveSoftware/Proton/files/2793527/error.log)
[output_log.txt](https://github.com/ValveSoftware/Proton/files/2793528/output_log.txt)

## System Information
- GPU: AMD Radeon 6990
- Driver/LLVM version: 3.1 Mesa 19.0.0-devel - padoka PPA
- Distro version: Ubuntu Mate 18.04 
- Link to full system information report as [Gist](https://gist.github.com/): https://gist.github.com/KGOrphanides/55e8ff25837ac5edc1fb7695c2a2f568
- Proton version: 3.16-6 beta

kisak-valve:
Hello @KGOrphanides, your Terascale 3 video card does not support vulkan, so your best bet is to add `PROTON_USE_WINED3D=1 %command%` or try some of the -force-renderer options from https://docs.unity3d.com/530/Documentation/Manual/CommandLineArguments.html.

KGOrphanides:
Thank you. Can confirm the game is now working.

PR #1749, opened 2018-10-15
Force wined3d if there's no Vulkan support

dreamer:
*Because I got annoyed by pasting PROTON_USE_WINED3D11=1 for every game on my old Thinkpad X220...*

Implementation queries if system supports any Vulkan physical devices. If vulkan library can be loaded and there is at least one physical device available, then proceed as normal - otherwise inject PROTON_USE_WINED3D11 variable to prevent dxvk from reporting no DirectX11 support at all.

According to https://store.steampowered.com/hwsurvey?platform=linux ~13-15% of Linux Steam users still do not have hardware support for Vulkan API (Intel Sandy Bridge or older, NVIDIA GTX 5xx series or older and AMD Radeon HD 76xx or older). Those systems are quite old, but often capable of running e.g. most Unity games.

Fixes e.g. Gorogoa (557600) #1711, Tesla Effect: A Tex Murphy Adventure (261510), and probably any other Unity title using only D3D11 renderer.

In future it could be expanded to filter devices providing incomplete or broken Vulkan support (and e.g. either report warnings in log or forcing wine3d311) to mitigate issues like broken fonts in #791.

### DIFF
--- a/vkquery.py
+++ b/vkquery.py
@@ -1,0 +1,99 @@
+#!/usr/bin/env python3
+
+# pylint: disable=wildcard-import, unused-wildcard-import, invalid-name
+
+"""Query Vulkan capabilities"""
+
+from ctypes import *
+
+VkResult = c_int32  # enum (size == 4)
+VK_SUCCESS = 0
+VK_ERROR_INITIALIZATION_FAILED = -3
+
+VkStructureType = c_int32  # enum (size == 4)
+VK_STRUCTURE_TYPE_APPLICATION_INFO = 0
+VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO = 1
+
+VkInstanceCreateFlags = c_int32  # enum (size == 4)
+
+VkInstance = c_void_p  # handle (struct ptr)
+
+
+def vk_make_version(major, minor, patch):
+    """
+    VK_MAKE_VERSION macro logic for Python
+
+    https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#fundamentals-versionnum
+    """
+    return c_uint32((major << 22) | (minor << 12) | patch)
+
+
+class VkApplicationInfo(Structure):
+
+    """Python shim for struct VkApplicationInfo
+
+    https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkApplicationInfo.html
+    """
+
+    # pylint: disable=too-few-public-methods
+
+    _fields_ = [('sType', VkStructureType),
+                ('pNext', c_void_p),
+                ('pApplicationName', c_char_p),
+                ('applicationVersion', c_uint32),
+                ('pEngineName', c_char_p),
+                ('engineVersion', c_uint32),
+                ('apiVersion', c_uint32)]
+
+    def __init__(self, name, version):
+        super().__init__()
+        self.sType = VK_STRUCTURE_TYPE_APPLICATION_INFO
+        self.pApplicationName = name.encode()
+        self.applicationVersion = vk_make_version(*version)
+        self.apiVersion = vk_make_version(1, 0, 0)
+
+
+class VkInstanceCreateInfo(Structure):
+
+    """Python shim for struct VkInstanceCreateInfo
+
+    https://www.khronos.org/registry/vulkan/specs/1.1-extensions/man/html/VkInstanceCreateInfo.html
+    """
+
+    # pylint: disable=too-few-public-methods
+
+    _fields_ = [('sType', VkStructureType),
+                ('pNext', c_void_p),
+                ('flags', VkInstanceCreateFlags),
+                ('pApplicationInfo', POINTER(VkApplicationInfo)),
+                ('enabledLayerCount', c_uint32),
+                ('ppEnabledLayerNames', c_char_p),
+                ('enabledExtensionCount', c_uint32),
+                ('ppEnabledExtensionNames', c_char_p)]
+
+    def __init__(self, app_info):
+        super().__init__()
+        self.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO
+        self.pApplicationInfo = pointer(app_info)
+
+
+def is_vulkan_supported():
+    """
+    Returns True iff vulkan library can be loaded, initialized,
+    and reports at least one physical device available.
+    """
+    vulkan = None
+    try:
+        vulkan = CDLL('libvulkan.so.1')
+    except OSError:
+        return False
+    app_info = VkApplicationInfo('vkinfo', version=(0, 1, 0))
+    create_info = VkInstanceCreateInfo(app_info)
+    instance = VkInstance()
+    result = vulkan.vkCreateInstance(byref(create_info), 0, byref(instance))
+    if result != VK_SUCCESS:
+        return False
+    dev_count = c_uint32(0)
+    result = vulkan.vkEnumeratePhysicalDevices(instance, byref(dev_count), 0)
+    vulkan.vkDestroyInstance(instance, 0)
+    return result == VK_SUCCESS and dev_count.value > 0
--- a/vkquery.py
+++ b/vkquery.py
@@ -77,23 +77,36 @@ class VkInstanceCreateInfo(Structure):
         self.pApplicationInfo = pointer(app_info)
 
 
-def is_vulkan_supported():
-    """
-    Returns True iff vulkan library can be loaded, initialized,
-    and reports at least one physical device available.
-    """
-    vulkan = None
-    try:
-        vulkan = CDLL('libvulkan.so.1')
-    except OSError:
-        return False
-    app_info = VkApplicationInfo('vkinfo', version=(0, 1, 0))
-    create_info = VkInstanceCreateInfo(app_info)
-    instance = VkInstance()
-    result = vulkan.vkCreateInstance(byref(create_info), 0, byref(instance))
-    if result != VK_SUCCESS:
-        return False
-    dev_count = c_uint32(0)
-    result = vulkan.vkEnumeratePhysicalDevices(instance, byref(dev_count), 0)
-    vulkan.vkDestroyInstance(instance, 0)
-    return result == VK_SUCCESS and dev_count.value > 0
+class VkQuery():
+    """Storage for queried Vulkan information"""
+
+    def __init__(self):
+        try:
+            self.lib = CDLL('libvulkan.so.1')
+        except OSError:
+            self.lib = None
+            self.dev_count = 0
+
+        self.dev_count = self.__enumerate_devices__()
+
+    def lib_found(self):
+        """Returns true iff Vulkan library was loaded."""
+        return self.lib is not None
+
+    def device_available(self):
+        """Returns number of physical devices supporting Vulkan API."""
+        return self.dev_count > 0
+
+    def __enumerate_devices__(self):
+        app_info = VkApplicationInfo('vkinfo', version=(0, 1, 0))
+        create_info = VkInstanceCreateInfo(app_info)
+        instance = VkInstance()
+        result = self.lib.vkCreateInstance(byref(create_info), 0, byref(instance))
+        if result != VK_SUCCESS:
+            return 0
+        dev_count = c_uint32(0)
+        result = self.lib.vkEnumeratePhysicalDevices(instance, byref(dev_count), 0)
+        self.lib.vkDestroyInstance(instance, 0)
+        if result != VK_SUCCESS:
+            return 0
+        return dev_count
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -272,8 +272,12 @@ $(PROTON37_TRACKED_FILES_TARGET): $(addprefix $(SRCDIR)/,proton_3.7_tracked_file
 USER_SETTINGS_PY_TARGET := $(addprefix $(DST_BASE)/,user_settings.sample.py)
 $(USER_SETTINGS_PY_TARGET): $(addprefix $(SRCDIR)/,user_settings.sample.py)
 
+VKQUERY_TARGET := $(addprefix $(DST_BASE)/,vkquery.py)
+$(VKQUERY_TARGET): $(addprefix $(SRCDIR)/,vkquery.py)
+
 DIST_COPY_TARGETS := $(TOOLMANIFEST_TARGET) $(FILELOCK_TARGET) $(PROTON_PY_TARGET) \
-                     $(PROTON37_TRACKED_FILES_TARGET) $(USER_SETTINGS_PY_TARGET)
+                     $(PROTON37_TRACKED_FILES_TARGET) $(USER_SETTINGS_PY_TARGET) \
+                     $(VKQUERY_TARGET)
 
 DIST_VERSION := $(DST_DIR)/version
 DIST_OVR32 := $(DST_DIR)/lib/wine/dxvk/openvr_api_dxvk.dll
--- a/proton
+++ b/proton
@@ -15,6 +15,7 @@ import sys
 import tarfile
 
 from filelock import FileLock
+from vkquery import is_vulkan_supported
 
 #To enable debug logging, copy "user_settings.sample.py" to "user_settings.py"
 #and edit it if needed.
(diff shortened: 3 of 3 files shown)


aeikum:
It's not clear if we want to do this, due to the user support fractal it implies. FWIW, you can specify PROTON_USE_WINED3D11 in the user_settings.py file so you don't have to do it for every game.

dreamer:
@aeikum but it still means, that game does not work out of the box... I literally see no downsides here - it only improves the situation for some people so they don't need to investigate why "DirectX11 is not supported" on their Linux system? Non-technical people are more likely to request refund from Steam than look for workarounds...

Jacalz:
Same issue for me, I think @dreamer is doing the right thing here 👍

sigmaSd:
I also really like like this pr, a lot of ppl report games as broken on [https://spcr.netlify.com/](https://spcr.netlify.com/) because they just don't know any technical stuff  , just one suggestion if a user have a vulkan capable gpu but vulkan-icd-loader is not installed this will still inject the wined3d11 workaround , I'm not sure if that is the best idea , probably the user should be informed to install the library instead.

dreamer:
Rebased on top of 3.16-2 and adapted to use new proton environment variable.

vkquery.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __init__(self, app_info):
+        super().__init__()
+        self.sType = VK_STRUCTURE_TYPE_INSTANCE_CREATE_INFO
+        self.pApplicationInfo = pointer(app_info)
+
+
+def is_vulkan_supported():
+    """
+    Returns True iff vulkan library can be loaded, initialized,
```
ghost:
```suggestion
    Returns True if vulkan library can be loaded, initialized,
```

dreamer: https://en.wikipedia.org/wiki/iff

tannisroot:
Can this check if a 32 bit instance can be created too? Sometimes people have Vulkan set up properly except they miss a 32 bit loader.

dreamer:
@tannisroot can you show an example when it's a problem? I've never experienced it myself.

@aeikum Seems like 3.16-4 included similar functionality - several games, that I tested that needed PROTON_USE_WINED3D now work out of the box (e.g. Tesla Effect (261510)). Did you implement this functionality by other means or just force-off dxvk for more titles in Steam compat database?

[edit] nope, I just changed my user_settings.py to manually disable dxvk on this machine and forgot about it. In result, I made several wrong Platinum reports on protondb.com, because I thought no workarounds were required, damnit :/

tannisroot:
Yes. If I remove lib32-vulkan-icd-loader, all games launch just fine and I don’t see a warning, but 32 bit dxvk games can’t work because of missing 32 bit vulkan loader.

cjwijtmans:
looks good to me.

dreamer:
Rebased on top of 3.16-4

Guy1524:
I don't think this is a very good idea, wined3d is much worse than the alternative, DXVK.  Users should only attempt it as a very last resort, and should try to get their Vulkan setups working correctly instead.  Also, if you have hardware that doesn't support Vulkan, you probably won't have have enough headroom to endure wined3d's awful performance.

keraldi:
A prompt asking the user whether or not they want to use wine3d would probably make it more feasible. Easy access to information through steam whether or not vulkan support is possible/vulkan is installed correctly is another thing.

dreamer:
@Guy1524 It's a fair point, I didn't think about users who have hardware capable of running Vulkan, but misconfigured system... although I don't think it applies here - steam runtime comes with it's own libvulkan.so and it's on user to have properly installed drivers (otherwise they won't be able to enjoy any game on Steam). This PR will (positively) affect users, whose hardware do not support DXVK. It's up to 15% of user base, as I mentioned earlier.
 
Maybe blacklist of hardware ids would work better than querying vulkan, but I am not sure.

Another way forward might be to implement this as opt-in to be configured through STEAM_COMPAT_CONFIG - it would be an optional switch to be decided per game title during whitelisting.

> Also, if you have hardware that doesn't support Vulkan, you probably won't have have enough headroom to endure wined3d's awful performance.

I don't agree at all - there are many non-demanding games, that work perfectly fine with wined3d3 even on older hardware. I wrote this patch with [Gorogoa](https://store.steampowered.com/app/557600/Gorogoa/) in mind, after seeing in forum users who bought the game to play using Proton, didn't know about proton env variables and requested a refund in result. I finished the game using wined3d and it worked just fine.

@Iorce 

> A prompt asking the user whether or not they want to use wine3d would probably make it more feasible.

I do not agree. If dxvk works, then there's no point in asking (just use dxvk). If it doesn't, then there's no point in asking (just use wined3d or let game report missing DX11 - current behaviour). And I leave information in log if fallback happened or not.

> Easy access to information through steam whether or not vulkan support is possible/vulkan is installed correctly is another thing.

Yes :( This info is missing from Steam System Information. Also, vulkan detector that comes bundled with Steam (`vulkandriverquery`) crashes on assert when there's no Vulkan support (issue was already reported long time ago).

ghost:
@Guy1524: I don't think that is correct. I use wine3d3 to play games like Starcraft 2 as well as Final Fantasy X before I'd even tried dxvk.  Plenty of older games work just fine with wined3d.  I'm not a big gamer, so I just try to play what works in my 2013 Haswell based laptop with an nvidia mobile card. Bumblebee can't work with Vulkan and even the Intel card spews out info saying that Vulkan isn't complete for Haswell.

Remeber that not all people who play games have the same needs as you.

Areiser:
@Guy1524 to chime in here, I was honestly surprised this was an issue. My graphics card does not support vulkan on mesa (R9 270, I know 2013 but still it plays most modern games for now) and I couldn't get a single game to start, even if it was whitelisted!
So I think for people new to steam play the possibility of using this cool new tool just to have a game not work at all (game is just not starting) is a bad situation. Also, a lot of laptop users with older intel integrated graphics might run into this problem when trying low-resource indie games.

doitsujin:
@Areiser Mesa does support Vulkan on your card, but requires the `amdgpu` kernel driver (it defaults to `radeon` on older AMD GPUs). You can enable it by adding `radeon.si_support=0 amdgpu.si_support=1` to your kernel command line.

kisak-valve:
There's a fairly well made write up for using vulkan with early GCN cards at #813.

FinixFighter:
I agree with dreamer! Same issue with multiple games. I think this can improve a lot satisfaction of gamers. There are people who just want to play. Some people will never go on github or make searches for solving the problem, they just want click and play. If the game starts when clicking "play", ok; but if it doesn't start, they will switch back to Windows (this happens for real, personal experience) XD

nsrosenqvist:
I play a lot of retro and 2D games on my older intel laptop. It's never going to get Vulkan support and I have to configure the launch options for every single game. There's already a warning when running a game in proton for the first time, why not just add a blurb there about Vulkan too? I had to search around to find out about that environment variable and I'm not sure everyone would. So enabling it by default definitely helps a lot

aeikum:
You don't have to set it for every game. You can set the variable globally by launching Steam with it already set, or add it to user_settings.py for each proton version. You'll also only need it for games that use d3d10 or d3d11. Older games that use d3d <=9 already use the GL backend.

mirh:
It still is something that should be automatic though?
You already detect other APIs versions in the steam client. Just add a check for vulkan and call it a day.

HomerSp:
If the automatic wined3d switch is unfeasible (I'm not sure it should be automatic, tbh), at least add a warning message and a toggle somewhere in the UI that the game will launch without vulkan. I was trying to find out earlier why a game crashed upon launching (using default proton settings), and the error message in the log just said that it failed to create a d3d11 device - I had to turn on debug messages to see that it failed to load required vulkan extensions.
As it turns out, I was missing the mesa vulkan drivers, and it's all working fine now. But it would have saved me some time if there was an easier way to tell if dxvk will be used or not, and why.

Couldn't it really be an option in Steam Play settings? And if the vulkan drivers are missing, or your graphics card doesn't support vulkan, show an error/warning message if you try to enable dxvk support?

codeman101:
I agree with this so much that I can't believe proton doesn't already do it. This needs to be added to the next milestone.

dreamer:
@codeman101 I see one reason why this change might be unfeasible in its current state: users, who have not installed properly their drivers and are missing Vulkan support in the result. For these users, it might be better if the game fails, prompting them to investigate (and maybe leave clearly visible info in the logfile).

However, I still see no reason why fallback to OpenGL shouldn't happen automatically for hardware, that definitely does not support Vulkan. I see two approaches here:

- Implement a PCI ID based blacklist. I started some work on this and it's harder than anticipated - NVIDIA does not provide OpenGL extension to read PCI ID (MESA conveniently does it), dealing with OpenGL and X in pure python without external libraries (only ctypes) is PITA and there is no reliable, semi-complete PCI ID database that I found, OR…
- Leave the code as it is, except change behaviour when Vulkan library was not found. I will rebase the patch and modify this PR in this manner later today.

mirh:
There's no need to keep a pciid list. 
Not having the vulkan ICD to begin with should appear quite different than having it, but no hardware support.

codeman101:
@dreamer 

I think a solution to the problem you mention is to have there be a popup if the game supports vulkan but the user doesn't have property installed there should be a popup telling them that they don't and what packages they are missing and then automatically revert to DX11.

Also if it's hard to make it be a GUI steam client looking popup then just have it be a console window. I'd offer to do it myself but I've never done wine/proton coding so it'd be a learning curve for me. (but going through the learning curve is on my to do list :) )

dreamer:
@codeman101 We don't know which game uses which graphics API ahead of time. And any solution involving GUI would need to be implemented in Steam client - not in Proton.

codeman101:
@dreamer  Regarding the GUI I know that's exactly why I mentioned using a console window instead.

Regarding the API given  what proton does (translating Windows calls to Linux) I'm a bit surprised that the API can't be detected. That's like saying I can translate language X to English but I can't tell if someone is speaking language X unless they tell me that they are.

mirh:
Wonder in which language they are gonna tell you which language they speak at all to begin with? :thinking: 

Just let knowledgeable people handle that.

dreamer:
@codeman101 API can be detected at runtime. It cannot be reliably detected *before* the game starts, therefore there's no way to inform the user "to run this specific application you need Vulkan" and supress that message for applications, that don't need it. You could do some static binary analysis or suspend

This PR is about detecting hardware that does not support Vulkan and gracefully falling back to OpenGL instead of letting the game fail because of no DirectX11 support.

codeman101:
@mirh
> Wonder in which language they are gonna tell you which language they speak at all to begin with? thinking

It doesn't sound like you understand the relationship I was making in the analogy.

> Just let knowledgeable people handle that.

Of course. As I said before I don't know how to do proton/wine coding specifically. I'm just stating what is common sense to me.

@dreamer

Thanks for explaining that it's a runtime issue. That makes more sense. It's a shame but at least now I understand why it's hard to do. 

@dreamer

Wait if it can be detected at runtime why not have the popup appear when the user tries to press play? 

> This PR is about detecting hardware that does not support Vulkan and gracefully falling back to OpenGL instead of letting the game fail because of no DirectX11 support.

I agree that it should do that fallback I was just trying to suggest a solution for letting the user know that they don't have vulkan installed.

dreamer:
Game runtime, not Proton runtime.

codeman101:
@dreamer 
> Game runtime, not Proton runtime.

I would've figured the two types of runtime were more interlinked but if they're not hmmm. That's a good question. Well thanks for the explanation. If I think of a solution I'll post back.

mirh:
Again.. you just have to try opening `libvulkan.so`, then "create the device". 
If you fail point one, warn the user (even though I really don't think this nuisance should be needed.. you can just print whatever API's being used in the terminal, and competent users should notice any problem), otherwise if you point two doesn't succeed use wined3d. 
Vulkaninfo probably does the same too.

dreamer:
@mirh I don't even need to create a device - I just use `vkEnumeratePhysicalDevices` (have you read the code?). It's not acceptable in its current form to Proton maintainers, so I wonder how to improve it.

One way of addressing "user support fractal" concern would be to be more strict about falling back to OpenGL. ATM I decide to fallback when vulkan lib can't be loaded OR there are no physical devices. I plan to change it to vulkan lib can be loaded AND there are no physical devices - this way Proton won't automatically switch to OpenGL for users who just don't have drivers installed.

dreamer:
Rebased to the latest version of `proton_4.2` branch. Changes:

- Adapt to the new build system
- Make the fallback detection more strict (see adc301d3b)
- Add basic Vulkan information to logfile

@aeikum I hope this version pushes PR towards merging. If you still consider this change undesirable, then I'll close it to avoid unnecesary discussion.

thaewrapt:
Not really sure if it helps your discussion, but actually Vulkan absence is already detected at Steam runtime level, some assert in it's code being failed during Steam client startup. So, theoretically, Proton could be told that it's lacking Vulkan support via some environment variable, command-line option, etc. from Steam client itself.

aeikum:
DXVK is the only supported renderer for d3d10 and d3d11. I think adding an automatic fallback to wined3d would mislead users into thinking it is a supported renderer.

dreamer:
@thaewrapt I know. Before going ahead and implementing basic Vulkan handling using ctypes I've spent some time investigating if I can reuse some already existing mechanism. Binary in question is `~/.local/share/Steam/ubuntu12_64/vulkandriverquery`. But this program crashes on assert (ValveSoftware/steam-for-linux#5423) - meaning Steam does not know if Vulkan is supported or not - maybe it does not crash always (I don't know), but this coredumping does not give proper information to Steam client itself. As far as I can tell (I guess) that program tries to initialize Vulkan library using bunch of extensions and then fails on assert instead of gracefully shut down with an error - while my implementation does not use any extensions at all (I just enumerate physical devices, nothing more).

By nature of being closed source, I can't fix it. Also, this program uses some internal Steam API to communicate and there's no documentation available, so I would need to spend time reverse-engineering it, which I consider a wasted effort.

Overall using Vulkan on such basic level straight from Python was not such a big deal - only ~110 lines of code :).

dreamer:
@aeikum I see… I think most people consider it a user option and not a debug option.

OK, so for people arriving at this PR at a later date: if you have the hardware that **does not** support Vulkan and are annoyed at putting `PROTON_USE_WINED3D=1` option for every game - define this variable in `user_settings.py` file (it will work for all games using that Proton version). If you want for it to work for every Proton version (including future ones), add:

    export PROTON_USE_WINED3D=1

to your `.bash_profile` file.

Officially, Proton supports only DXVK as DX10/11 renderer. I am closing this PR - if we'll ever need the code, it's still available in history or on my branch (https://github.com/dreamer/Proton/commits/po/vkquery).

kisak-valve:
Stuff in `.bash_profile` is typically not applied to a gui-only session.

dreamer:
@kisak-valve According to [bash manual](https://linux.die.net/man/1/bash):

      ~/.bash_profile
              The personal initialization file, executed for login shells

Freshly logged in to Fedora to make sure it works in purely GUI environment (maybe different distros override this behaviour somehow?).

I stand by my comment ;)

mirh:
> I think adding an automatic fallback to wined3d would mislead users into thinking it is a supported renderer.

Currently there are users that [are](https://github.com/ValveSoftware/Proton/issues/1094) mislead into thinking proton/steam is just broken because the thing fails. 
The priority should be "people getting things done", rather than "signaling support". If whatever the "uncared for" reports start to flock, they should be pretty easy to detect, or dismiss if even.

James-E-A:
> AMD Radeon HD 76xx or older

It may be interesting to note that this was inaccurate when it was posted, and it's still somewhat inaccurate.

Graphics Core Next 1 (HD 7730–7990, 8570–8990; RX 240–280, 330–370, 430–450, 520) only got Vulkan support added to Ubuntu LTS in January of 2020, when they accepted Mesa version 20 with the RADV backend.

And Ubuntu LTS releases _still_ don't ship with functioning AMDGPU drivers for Graphics Core Next 1–3 cards out-of-the-box; users have to twiddle around with `si_enable` and `cik_enable` module parameters to get it working.

-----

someone could very well have a somewhat recent GPU and _still_ get hit by this due to Canonical being slow to enable upstream graphics driver updates.

PR #1763, opened 2018-10-17
 Enabled logging for non-appid invocation

ghost:
Allows Proton to still write a log file when not using an AppId number.  This allows debugging when, for example, Proton is used to run cmd.exe or any non-Steam game binary.  

Also reduces ambiguity about what is creating the log file and why by  changing the filename generation from ```steam-<appid>.log``` to ```steam-proton-<appid>.log```, or in the case of non-appid invocation, ```steam-proton-unknown.log```.  Since Steam does not do similar logging for native apps, it seems prudent to specify that this is a Proton/Wine log file and not a file that Steam would create under other circumstances.

### DIFF
--- a/proton
+++ b/proton
@@ -239,12 +239,16 @@ else:
 lfile = None
 if "SteamGameId" in env:
     if env["WINEDEBUG"] != "-all":
-        lfile_path = os.environ["HOME"] + "/steam-" + os.environ["SteamGameId"] + ".log"
+        lfile_path = os.environ["HOME"] + "/steam-proton-" + os.environ["SteamGameId"] + ".log"
         if os.path.exists(lfile_path):
             os.remove(lfile_path)
         lfile = open(lfile_path, "w+")
 else:
-    env["WINEDEBUG"] = "-all"
+    if env["WINEDEBUG"] != "-all":
+        lfile_path = os.environ["HOME"] + "/steam-proton-unknown" + ".log"
+        if os.path.exists(lfile_path):
+            os.remove(lfile_path)
+        lfile = open(lfile_path, "w+")
 
 prefix_lock = FileLock(os.environ["STEAM_COMPAT_DATA_PATH"] + "/pfx.lock", timeout=-1)
 with prefix_lock:


proton
```diff
@@ -239,12 +239,16 @@ else:
 lfile = None
 if "SteamGameId" in env:
     if env["WINEDEBUG"] != "-all":
-        lfile_path = os.environ["HOME"] + "/steam-" + os.environ["SteamGameId"] + ".log"
+        lfile_path = os.environ["HOME"] + "/steam-proton-" + os.environ["SteamGameId"] + ".log"
```
dreamer, line 242: If you want to change the default name of a log file, then you need to update the documentation in `README.md` as well.

proton
```diff
@@ -239,12 +239,16 @@ else:
...
         if os.path.exists(lfile_path):
             os.remove(lfile_path)
         lfile = open(lfile_path, "w+")
 else:
-    env["WINEDEBUG"] = "-all"
```
dreamer, line 247: I think you intentionally removed turning on '-all' here... Can you comment on that?

If we remove it from here, then maybe it makes sense to get rid of `env["WINEDEBUG"] != "-all"` as well?

proton
```diff
@@ -239,12 +239,16 @@ else:
...
+    if env["WINEDEBUG"] != "-all":
+        lfile_path = os.environ["HOME"] + "/steam-proton-unknown" + ".log"
+        if os.path.exists(lfile_path):
```
dreamer, line 249: I think there's no need to duplicate lines 243-245 (to 249-251). You could move `lfile_path` variable to separate scope and avoid some code duplication this way:

    lfile_path = os.environ["HOME"] + "/steam-proton-unknown.log"
    appid = env.get("SteamGameId")
    if appid:
        lfile_path = os.environ["HOME"] + "/steam-proton-" + appid + ".log"
    if env["WINEDEBUG"] != "-all":
        # rest of logic follows

Also, it might be good idea to introduce `os.path.join` to proton file finally ;)

dreamer (changes requested):
I was bitten by lack of logs as well - if this change was merged I could simplify my personal wrapper script for testing proton, so :+1: :)

Left some suggestions in review (documentation update is rather necessary).

dreamer:
Oh, also - how about rebasing your patch on top of 3.16-2 to avoid merge commit in PR?

proton
ghost: The ```-all``` here effectively overrides and disables all user debugging, even when specifically enabled in ```user_settings.py```.  This change allows the same debug flags to pass through as if the script were running an AppId invocation, thus making both use cases consistent in regard to logging.

You're right, the ```env["WINEDEBUG"] != "-all"``` line may be unnecessary, but I'd have to take a closer look at it.  I'm not much of a Python guy, but I can take a peek at that cleanup later today.

proton
ghost: Yeah, my bad, should have proposed doc update too.  Are we okay with renaming these files slightly then?

proton
ghost: There's probably a good bit of cleanup to do : )  Might be better to wrap that up in a separate pull for the sake of scope?
I'll try to take a look at that later on today.

aeikum:
Hmm, it's a tricky problem.

It's definitely nice to have logging when we don't have an appid (e.g. so we can capture +server logs, and logs during installscript processing). But, the log file being called "unknown" and also separate from the game's log is rather ugly. It would be nicer to have them interleaved. You also have the problem that the unknown log will be wiped every time Steam launches something without an appid, which it will do several times during the game launch process, so you only capture whatever happens to be last run without an appid.

What I normally do in this situation is change all logging to go to ~/steam.log, which we don't delete and instead open in append mode. This has the disadvantage of appending logs to the same file endlessly, leading to a huge log file if you forget to delete it between runs.

I'd maybe be open to some switch to provide the latter functionality easily. But I think the proposed solution here is a little too clunky.

dreamer:
Hmm, we can access name of executable in sys.argv, so it can be included in log file instead of 'unknown'. Perhaps after normalization (e.g. lowercase and replace spaces by minuses).

e.g. for appid [260230](https://store.steampowered.com/app/260230/) executable is called "Valiant Hearts.exe", so log would be called: `steam-proton-valiant-hearts.log`.

proton
dreamer: I think it's appropriate to keep it in this PR, but it's ok to put it in new commit (or you might want to amend or rebase --interactive).

aeikum:
I added a switch so if you have `PROTON_LOG_ALL_PROCESSES` set, then it will log everything to `~/steam-proton.log` without deleting the file (i.e. append endlessly). This lets you gather logs while it is configuring the prefix, and also grab wineserver and service logs, which is really useful for controller debugging. It will also log even if the SteamAppId environment variable is unset.

The big disadvantage here is that we append endlessly... I'm not sure how we could determine when to erase the log (i.e. detect the invocation that we actually want to log vs earlier invocations of installscript handling stuff).

aeikum:
I pushed that to the proton_4.2-log-all branch here. Feedback welcome. https://github.com/ValveSoftware/Proton/tree/proton_4.2-log-all

aeikum:
This is possible now in both 6.3 and 7.0 series by changing the `append_forever` parameter to the `setup_logging` funciton. For example, apply this diff to 7.0-1:

````diff
diff -ur a/proton b/proton
--- a/proton	2022-02-16 08:36:15.434437741 -0600
+++ b/proton	2022-02-16 08:36:24.587854378 -0600
@@ -1130,7 +1130,7 @@
             self.env["WINE_CRASH_REPORT_DIR"] = self.env["PROTON_CRASH_REPORT_DIR"]
 
         if self.env["WINEDEBUG"] != "-all":
-            if self.setup_logging(append_forever=False):
+            if self.setup_logging(append_forever=True):
                 self.log_file.write("======================\n")
                 with open(g_proton.version_file, "r") as f:
                     self.log_file.write("Proton: " + f.readline().strip() + "\n")
````

This will append to a log file in your Home directory called "steam-proton.log". It will never delete this file, so it will gather all Proton logging from all installscript and other non-game process, etc. However it will fill your disk up if you forget you have it enabled. It's intended as a temporary debugging utility.

The proton script will be restored to its original version whenever Steam updates Proton.

PR #1877, opened 2018-11-01
Add vagrant script for build vm.

madewokherd:
I have limited time, so I wanted to at least get this sent while I have some time and now that I know the build succeeds. This adds a Vagrantfile which automates the process of making a build vm and streamlines its use. There are unfortunately some important tasks remaining that I'm not able to finish at the moment:

- I had to remove the dxvk version marker for this case because it requires .git folders which I decided not to sync into the vm. This may need more thought if it's important.
- The readme should really have documentation of how to create and use the vm.
- I should test that the build created by the vm actually works.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,44 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+# Vagrant file for setting up a build environment for Proton.
+
+Vagrant.configure(2) do |config|
+  config.vm.box = "generic/debian9"
+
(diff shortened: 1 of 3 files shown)


aeikum:
I'm fine with dropping the DXVK version file.

I'll go read up on Vagrant a little, then try it out and see if it builds for me. If so, I don't see any reason not to merge this. We can refine it going forward.

aeikum:
Thanks, this was merged in.

PR #1925, opened 2018-11-05
[DO NOT MERGE] FAudio support

flibitijibibo:
This is a branch that adds FAudio and uses a branch of Wine that integrates FAudio as the XAudio2 implementation.

This is NOT meant to be merged; FAudio will need to be integrated into a future Proton release, but for the purposes of testing, this is based on 3.16.

Unofficial TODO:
- [ ] Get distributors to package FAudio (probably starting with 19.01)
- [ ] Get Wine patches into upstream (or more likely, wine-staging, since this is a total rewrite)
- [ ] Update Proton's Wine version
- [ ] Integrate the patches in this PR

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "wine"]
     path = wine
-    url = https://github.com/ValveSoftware/wine/
+    url = https://github.com/flibitijibibo/wine/
 [submodule "dxvk"]
 	path = dxvk
 	url = https://github.com/ValveSoftware/dxvk/
@@ -19,3 +19,6 @@
 [submodule "fonts/liberation-fonts"]
 	path = fonts/liberation-fonts
 	url = https://github.com/liberationfonts/liberation-fonts
+[submodule "FAudio"]
+	path = FAudio
+	url = https://github.com/FNA-XNA/FAudio/
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -176,6 +176,10 @@ FFMPEG_OBJ64 := ./obj-ffmpeg64
 FFMPEG_CROSS_CFLAGS :=
 FFMPEG_CROSS_LDFLAGS :=
 
+FAUDIO := $(SRCDIR)/FAudio
+FAUDIO_OBJ32 := ./obj-faudio32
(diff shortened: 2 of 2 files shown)


flibitijibibo:
~For anyone attempting to update to upstream FAudio...~

~https://github.com/flibitijibibo/wine/pull/1#issuecomment-437537624~

~Will try to fix this on Monday.~

flibitijibibo:
Big update to FAudio (and Wine as a result), fixes a bunch of games including Mega Man 11, Banished, and probably a bunch of others.

flibitijibibo:
One more big update that fixes issues with callbacks - games that use OnBuffer* and On*Processing* should be fixed by this update.

flibitijibibo:
Quick update to fix a very particular multi-device bug in FAudio. Proton is now using the `devices` branch, since the patch on that branch is very important and is perfectly fine for Proton's platform(s). Not sure when we'll be able to fix it for the rest though:

https://github.com/FNA-XNA/FAudio/pull/72

flibitijibibo:
One more update, this fixes some threading issues that Andrew found and also fixes 1024-byte MSADPCM blocks. Puyo Puyo Tetris is one example that should work now.

flibitijibibo:
Latest update sets X3DAudio and XAPOFX to prefer builtin instead of native. This is entirely a guess but I think this may fix some of the buzzing that's been going on for certain games using XAPOFX. X3DAudio is also good enough to use by default, I think... the DELAY flag isn't very commonly used so I'm not too worried about that.

flibitijibibo:
Minor update to FAudio to log the everloving crap out of every API/function's enter and exit behavior, as well as every mutex action. There are also a couple XACT fixes for anyone trying this out with RE4...

flibitijibibo:
One more FAudio update, we're back on master and there's a XAPO fix that cleans up the audio for Hellblade and probably lots of other games (UE4 in particular).

flibitijibibo:
The latest FAudio moves to CMake, so it’s no longer compatible with the build changes in this patch. I’d make an updated patch, but uh... well, maybe just wait a bit longer...

PR #1931, opened 2018-11-06
Shellcheck fixes for configure.sh and steamrt-bootstrap.sh

Kedstar99:
In steamrt-bootstrap,sh
  + Used quotes to prevent word splitting SC2046
  + Used $() notation rather than legacy backtick SC2006
  + which is non-standard. 'command -v' is builtin SC2230

In configure.sh
  + Assigned to local variable separately to avoid masking return values SC2155
  + Used to quote to prevent glob matching SC2053
  + Used -z command rather than ! -n SC2236

Things still to fix according to shellcheck:

Line 5:
SRCDIR="$(dirname "$0")"
^-- SC2034: SRCDIR appears unused. Verify use (or export if used externally).
 
Line 95:
  echo >> "$MAKEFILE" "# Generated by: $THIS_COMMAND"
  ^-- SC2129: Consider using { cmd1; cmd2; } >> file instead of individual redirects.
 
Line 106:
  echo >> "$MAKEFILE" "STEAMRT64_MODE  := $(escape_for_make "$steamrt64_type")"
  ^-- SC2129: Consider using { cmd1; cmd2; } >> file instead of individual redirects.
 
Line 129:
invalid_args=""
^-- SC2034: invalid_args appears unused. Verify use (or export if used externally).

I am a little curious about the 'echo >> commands'. Could I know what they are doing?


aeikum:
Adding quotes around most variables seems like a good idea, but I'm a little confused why it complains about quotes in some areas but not others. Seems like it missed quite a few.

As for the rest of it, I'm not really sure what the benefit is. I don't see anything wrong with backticks and which, and both scripts specify bash in the hashbang, so whatever. I guess I'm not opposed, but I also don't see the benefit.

@Nephyrin Any thoughts?

Nephyrin:
👍

While some of these are more fixes-for-the-linter-god rather than real issues, they seem fine to implement.  None are really impacting readability.

The [SC2053] fixes might be actual bugs ("You don't need quotes in `[[ ]]` except when you do")

[SC2053]: https://github.com/koalaman/shellcheck/wiki/SC2053

Kedstar99:
Hello again,

I was just playing around and I think I fixed SC2129 too which is a minor style issue.

Minor performance gain too.

If you want me to squash both commits and amend the original, I am happy to do so.

I tested the echo changes in a minor text shell script and it seems to stay the same.

dreamer:
@aeikum 
> I don't see anything wrong with backticks

Here is handy list of reasons why backticks should be avoided:

[SC2006](https://github.com/koalaman/shellcheck/wiki/SC2006):
> Backtick command substitution \`...\` is legacy syntax with several issues.
>
> * It has a series of undefined behaviors related to quoting in POSIX.
> * It imposes a custom escaping mode with surprising results.
> * It's exceptionally hard to nest.
>
> $(...) command substitution has none of these problems, and is therefore strongly encouraged.

aeikum:
Cool, thanks. Will merge soon.

aeikum:
Thanks, this was merged as 416bea727e43c9dd1bb744d9c62f2903e8ea091c.

PR #1988, opened 2018-11-17
Proton log section in Compat Reports

cybik:
It's really just clarification


aeikum:
Any thoughts, @kisak-valve ?

kisak-valve:
I have no particular opinion one way or another for this change. While having a Proton log in the compatibility reports is often a helpful starting point, I don't think it's strictly mandatory to get a discussion started.

aeikum:
I think things have been going OK here.

PR #2075, opened 2018-12-10
vrclient: Add IVRCompositor_017

hdmap:
I'm still not sure how to go about this so feel free to tell me I'm way off the mark here. The Lab requires IVRCompositor_017 but as far as I can tell it was never made public with a header as IVRCompositor went from 016 to 018 from openvr 1.0.3 to 1.0.4. With these changes I can get the game to start and some of the games in it work but everything else causes the game to hang; I'm not sure if those are unrelated problems or symptoms of something I've missed here.

### DIFF
--- a/vrclient_x64/gen_wrapper.py
+++ b/vrclient_x64/gen_wrapper.py
@@ -27,6 +27,7 @@ sdk_versions = [
     "v1.0.5",
     "v1.0.4",
     "v1.0.3",
+    "v1.0.3a",
     "v1.0.2",
     "v1.0.1",
     "v1.0.0",
@@ -97,7 +98,6 @@ aliases = {
     #is compatible.
 #    "SteamClient012":["SteamClient013"],
 #    "SteamUtils004":["SteamUtils003"], # TimeShift uses SteamUtils003
-
 
     #leaving these commented-out. let's see if they turn up in practice,
     #and handle them correctly if so.
@@ -171,7 +171,7 @@ def ivrcompositor_post_present_handoff(cppname, method):
     return "ivrcompositor_post_present_handoff"
 
 def ivrcompositor_wait_get_poses(cppname, method):
-    for version in ["016", "018", "019", "020", "021", "022"]:
+    for version in ["016", "017","018", "019", "020", "021", "022"]:
         if version in cppname:
             return "ivrcompositor_wait_get_poses"
     return None
@@ -564,7 +564,8 @@ WINE_DEFAULT_DEBUG_CHANNEL(vrclient);
     constructors.write("    {\"FnTable:%s\", &create_%s_FnTable, &destroy_%s_FnTable},\n" % (iface_version, winclassname, winclassname))
     if iface_version in aliases.keys():
         for alias in aliases[iface_version]:
-            constructors.write("    {\"%s\", &create_%s}, /* alias */\n" % (alias, winclassname))
+            constructors.write("    {\"%s\", &create_%s, &destroy_%s}, /* alias */\n" % (alias, winclassname, winclassname))
+            constructors.write("    {\"FnTable:%s\", &create_%s_FnTable, &destroy_%s_FnTable},\n" % (alias, winclassname, winclassname))
 
     generate_c_api_thunk_tests(winclassname, methods, method_names)
 
--- a/vrclient_x64/gen_wrapper.py
+++ b/vrclient_x64/gen_wrapper.py
@@ -171,7 +171,7 @@ def ivrcompositor_post_present_handoff(cppname, method):
     return "ivrcompositor_post_present_handoff"
 
 def ivrcompositor_wait_get_poses(cppname, method):
-    for version in ["016", "017","018", "019", "020", "021", "022"]:
+    for version in ["016", "017", "018", "019", "020", "021", "022"]:
         if version in cppname:
             return "ivrcompositor_wait_get_poses"
     return None
--- a/vrclient_x64/openvr_v1.0.3a/ivrclientcore.h
+++ b/vrclient_x64/openvr_v1.0.3a/ivrclientcore.h
@@ -1,0 +1,36 @@
+//========= Copyright Valve Corporation ============//
+
+#include "openvr.h"
+
+namespace vr
+{
+
+class IVRClientCore
+{
+public:
+	/** Initializes the system */
+	virtual EVRInitError Init( vr::EVRApplicationType eApplicationType ) = 0;
+
+	/** cleans up everything in vrclient.dll and prepares the DLL to be unloaded */
+	virtual void Cleanup() = 0;
+
+	/** checks to see if the specified interface/version is supported in this vrclient.dll */
+	virtual EVRInitError IsInterfaceVersionValid( const char *pchInterfaceVersion ) = 0;
+
+	/** Retrieves any interface from vrclient.dll */
+	virtual void *GetGenericInterface( const char *pchNameAndVersion, EVRInitError *peError ) = 0;
+
+	/** Returns true if any driver has an HMD attached. Can be called outside of Init/Cleanup */
+	virtual bool BIsHmdPresent() = 0;
+
+	/** Returns an english error string from inside vrclient.dll which might be newer than the API DLL */
+	virtual const char *GetEnglishStringForHmdError( vr::EVRInitError eError ) = 0;
+
+	/** Returns an error symbol from inside vrclient.dll which might be newer than the API DLL */
+	virtual const char *GetIDForVRInitError( vr::EVRInitError eError ) = 0;
+};
+
+static const char * const IVRClientCore_Version = "IVRClientCore_002";
+
+
+}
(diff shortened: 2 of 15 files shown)


aeikum:
Hey, thanks! The way we've solved this for other interfaces is with the "aliases" list, near the top of gen_wrapper.py. However, we need to verify that the interface ABIs are actually binary compatible, and that (usually) requires some hints from the OpenVR developers, Valve.

Looking at the commit that changed IVRCompositor from version 16 to version 18 (ValveSoftware/openvr@84e877f1e099173378759189609b90870f422130), I noticed that IVRSystem also jumped from 12 to 14, so that may also affect The Lab.

So looks like we need Valve to verify that IVRCompositor 17 is compatible with version 18, and that IVRSystem version 13 is compatible with version 14, and then we can add them to the aliases list.  Or, we need them to release those undocumented versions, so we can support them correctly.

kisak-valve:
@JoeLudwig, @Plagman, is this something one of you would be interested in checking?

vrclient_x64/gen_wrapper.py
```diff
@@ -171,7 +171,7 @@ def ivrcompositor_post_present_handoff(cppname, method):
     return "ivrcompositor_post_present_handoff"
 
 def ivrcompositor_wait_get_poses(cppname, method):
-    for version in ["016", "018", "019", "020", "021", "022"]:
+    for version in ["016", "017","018", "019", "020", "021", "022"]:
```
JoeLudwig: No idea how picky we are about whitespace, but a space is missing here.

vrclient_x64/openvr_v1.0.3a/ivrclientcore.h
```diff
@@ -0,0 +1,36 @@
+//========= Copyright Valve Corporation ============//
```
JoeLudwig, line 1: The version number in this path looks suspiciously like an incorrect version number for the latest OpenVR SDK. If that's what it's meant to be the number should be right: 1.1.3b

vrclient_x64/vrclient_x64/struct_converters_103a.cpp
```diff
@@ -0,0 +1,100 @@
+#include <stdlib.h>
```
JoeLudwig, line 1: If the name of this file is also an SDK version number space should be left for two digits in the "build" number. We happened to only have three versions after we introduced version numbers of this style, but a full beta cycle will definitely result in double digits.

JoeLudwig: Oh, and the SDK version is wrong here too.

JoeLudwig:
Added some comments. I don't grok the actual thunking code, so I'll leave that for @Plagman 

My basic question is this: Where does the number 1.0.3 come from?

aeikum:
@JoeLudwig Yeah the approach taken by this PR is wrong. It's more about how we should support some undocumented interfaces that are used by The Lab. See my comment above.

JoeLudwig:
> So looks like we need Valve to verify that IVRCompositor 17 is compatible with version 18, and that IVRSystem version 13 is compatible with version 14, and then we can add them to the aliases list. Or, we need them to release those undocumented versions, so we can support them correctly.

What do you mean by "compatible"?  IVRCompositor gained a new function: IsMotionSmoothingEnabled. All the other calls are unchanged. Both 17 and 18 should work if fetched from the runtime.

IVRSystem is 19, which hasn't changed in a while...

aeikum:
Basically, can we give The Lab an IVRCompositor_018 object instead of the IVRCompositor_017 object that it requested?

For example, if 017's vtable has methods (A, B, C, D), but 018's vtable has (A, B, X, C, D), then when the game calls method C, it will actually be calling method X, and things will go badly.  However, if 018 has a compatible vtable layout and function signatures, then there is no problem.

Since 017 was skipped in the public header releases, it's impossible for me to answer this question.

Same applies for IVRSystem 012 -> 014. It skips 013.

hdmap:
I originally tried doing this with the aliasing approach (and looking back in the PR again I accidentally left some changes in gen_wrapper.py related to getting aliasing working for me). It looked to me like the type of scenario you just described and the game would crash pretty quickly that way so I assumed that they weren't binary compatible.

I totally missed IVRSystem. I never saw any complaining about it in the logs like I did with IVRCompositor but I can look again more closely later.

JoeLudwig:
Got it. You should be able to determine this yourself by diffing the public header with v17 and v18 (or v12 and v14 of IVRSystem). Finding very old versions like these might be easier from our side because we can diff the original ivrsystem.h and ivrcompositor.h instead of the combined openvr.h header.

Looking at IVRSystem on our side I see that several signatures changed from v12 to v14. No new functions were added, but if The Lab (or any other app?) calls GetProjectionMatrix or one of the other changed functions it's not going to be compatible.

For IVRCompositor 17->18, two functions were added in the middle, so they're not compatible.

We generally don't bump the version number for changes that just add new functions to the end of the vtable. It causes us to need an interface adapter in vrclient, which is mildly annoying to add. If the version number changes it's likely because the signature of an existing function changed, a function was deleted, or a function was added in the middle of the vtable.

aeikum:
Right, the trouble is there is no public header with version 17, so I have to ask you guys to do it. The commit in the openvr repo (mentioned above) goes directly from 16 to 18.

If 17 is not compatible with 18, then I'm not sure what our best option is... It'd be nice if you could release the SDK version that The Lab was built against (containing v17). If you can't do that, then maybe we could re-create it if you can give us the 16->17 diff.

Again, ditto for IVRSystem_013.

kisak-valve:
Since the interface was not made public, it's possible The Lab may be the only user of IVRCompositor_017. Maybe it would be less hassle to marginally bump the openvr header it's built against so it uses IVRCompositor_018 and rebuild?

JoeLudwig:
Ah, ok. Got it.

Releasing an SDK with two year old versions of headers would likely cause more confusion than it solved. I've attached a rough estimate of what the header from such an SDK might have been. Hopefully that helps?
[openvr.h.txt](https://github.com/ValveSoftware/Proton/files/2668667/openvr.h.txt)

We're working on Knuckles support in The Lab, so that'll bump the SDK it builds against up to something modern. That's not going to happen right away

aeikum:
Thanks, that will work fine. If that is what The Lab was built against, then we can just forget IVRSystem_013 until it actually turns up in a real game.

@hdmap If you want to work on this, I think you could synthesize an SDK by copying either the 1.0.3 or 1.0.4 SDK into a new dir (1.0.3x or something), and then dropping that openvr.h into it. A diff from 1.0.3 to the new version should look sane. This is pretty much what you already did, but with the real v17 in place.

Or you can leave it with me and I'll try to fix it some time.

JoeLudwig:
@aeikum It's difficult to know exactly what The Lab was built against. Since there was no released SDK somebody was copying headers around behind the scenes.

hdmap:
I'll give it a shot. Thanks!

aeikum:
@hdmap If you get a minute, it'd be nice to verify what (if any) IVRSystem interfaces The Lab requests. Should turn up in a +steamclient log if it requests any.

hdmap:
@aeikum Looks like it requests 011 and 014.

hdmap:
With this new header the Xortex game seems to work, although I can't figure out why it would now and not before.

The signature of ComputeDistortion also seems to have changed between this and 1.0.4 even though both are versioned as part of IVRSystem_014. vrclient is still being built with the 1.0.4 version.

aeikum:
Thanks @hdmap, that looks great. I was able to get in game with that patch. Trying to launch the Robot Repair demo crashed on a call to IVRApplications::LaunchInternalProcess, which looks like a bit of a bear.

If you want to keep working on this, probably that should be overriden via method_overrides with a custom implementation in vrclient_main.c. However, I'm not sure what it should do... I guess call the Linux implementation with a path to Wine in the first arg and rearrange the remaining args to insert the win32 binary path as the first arg. Not sure where we would get the Wine binary path off the top of my head. The working directory path will also need to be converted (see recent work in lsteamclient).

Regardless, I'll be merging your PR here for the next version of Proton, thanks!

JoeLudwig:
Two things:
* Robot Repair uses Source2 and a totally different build process than the rest of the lap, so it may have its own special interface versions and issues with those. You can probably just run vr.exe out of <lab install>/bin/win64 with whatever arguments are being passed to LaunchInternal process and make sure it's running in isolation before you try to get transitions in and out working.
* IVRApplications::LaunchInternalProcess attempts to let exe A start exe B while keeping the same app key and overlays active through the transition. The overlay bits are probably fine, but the bits where vrserver actually starts the other exe might be more problematic. Let me know if you run into trouble here. I can help with what's happening in vrsrever.

hdmap:
I just now realized this got merged in, so I'll close the PR. After an update to The Lab it's been hanging at 100% cpu for me when trying to load anything and I haven't pinpointed why or downloaded an old version to verify that the update is what changed things.

PR #2082, opened 2018-12-13
Enable ffmpeg by default and build FAudio with it

GloriousEggroll:
It was suggested that I make a PR with the necessary changes for faudio to be built with ffmpeg in proton and enabled in wine, so here we are.


flibitijibibo:
A possible optimization would be to remove FFmpeg from this repo and link to the Steam Runtime’s copy instead. The licensing limitations are the same for both projects, that would just make it so we’re building one less thing in Proton.

I’d also need to tweak FAudio a bit to use WMAERROR even when HAVE_FFMPEG is defined; not difficult but would make the user experience nicer when it’s not there.

mat8913:
The Steam Runtime seems to have a super old version of Libav (a fork of FFMpeg). I'm not sure if FAudio will be able to link to it.

GloriousEggroll:
changed to current branch and cleaned up conflicts, + added ffmpeg to dist: target

soredake:
Any progress?

flibitijibibo:
@soredake Have the patents expired yet?

foresto:
By the way, which patents are they?

flibitijibibo:
IANAL but I think some of them fall under the VC-1 bundle, but the complete list is unknown.

ASF expired a few years ago though! https://patents.google.com/patent/US6041345

EDIT: Was looking at the application status, which is always today...

foresto:
WMA v2 is old. It looks like the decoder in ffmpeg is [over 17 years old](https://github.com/FFmpeg/FFmpeg/commit/bc8d1857fbd8bd6af94ed8f164fa92cec51fb776), and I would expect any patents on the process or format to be even older. Aren't US patents limited to 20 years from the filing date?

Is there someone in the community who knows which patents make WMA decoding unsuitable for inclusion in Wine? It would be a pity to reject a useful audio decoder based on speculation or FUD. At the very least, knowing which patents we're afraid of would allow planning a date for making progress.

tgurr:
I'm by no means an expert, but I found https://wiki.multimedia.cx/index.php/Windows_Media_Audio mentioning https://patents.google.com/patent/US20050015246A1/en?oq=20050015246 and https://patents.google.com/patent/US20050015259A1/en?oq=20050015259 which both seem to be still valid until early 2026.

lieff:
@tgurr Both seems related to audio encoder algorithm, not decoder.

foresto:
> @tgurr Both seems related to audio encoder algorithm, not decoder.

Also, ffmpeg's WMA decoder (including WMAv2, used for xWMA games) already existed for several years before the dates on those patents. Another reason to doubt that they apply here.

tgurr:
@lieff @foresto Thank you very much for the clarification and sorry for the noise.

soredake:
Debian/ubuntu ship ffmpeg and faudio with ffmpeg support enabled.

Galcian79:
I have an issue with ffmpeg proton builds in Skyrim SE. No sound is played except the intro logo. It works fine in official proton with xaudio and the required hacks.
Tested with Proton 5.8 GE-MF-1 and proton 5.0.7

flibitijibibo:
FAudio now only needs `-DGSTREAMER=ON` to be configured with support for WMA. It's still up to the distributor to provide a GStreamer plugin, and we'll fail the same way as before if it's not available, but now we don't require FFmpeg to be present to build/run.

GloriousEggroll:
Closing this as the functionality has now been replaced with gstreamer

Issue #2058, opened 2018-12-04
The Misadventures of P.B. Winterbottom (40930)


flibitijibibo:
# Compatibility Report
- Name of the game with compatibility issues: The Misadventures of P.B. Winterbottom
- Steam AppID of the game: 40930

## System Information
- GPU: GTX 770
- Driver/LLVM version: 415.18
- Kernel version: 4.19.4-200.fc28
- Link to full system information report: [Gist](https://gist.github.com/flibitijibibo/3ecd7f85f35fcdfe724fa5c1c477cec2)
- Proton version: flibitProton, 3.16-4 Beta with the FAudio changes

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

[steam-40930.log](https://github.com/ValveSoftware/Proton/files/2646123/steam-40930.log)

## Symptoms 
Winterbottom has two paths, a D3D10 path and a D3D9 path. With the current Proton build, the game picks D3D10 but immediately hits a snag with Effects:

```
wine: Call from 0x7b43c6bd to unimplemented function d3dx10_43.dll.D3DX10CreateEffectFromResourceW, aborting
```

This is unimplemented in both Wine and DXVK. AFAIK D3D9 should work just fine, as the Effects for 9 are just SpriteBatch-like shaders and Wine's current implementation should be able to handle it. (And if not, this may be a good time to look at my implementation in MojoShader...)

Note that the game uses XACT for audio; this is known to work with FAudio but may not work with stock Wine.

## Reproduction
1. Boot game
2. Wait
3. Read logs

flibitijibibo:
After applying #2107, it does indeed look like Wine's d3dcompiler can't compile fx_2_0:

```
376373.176:0008:0009:fixme:d3dcompiler:compile_shader Compilation target "fx_2_0" not yet supported
376373.177:0008:0009:fixme:d3dx:d3dx9_effect_init Failed to parse effect, hr 0x8876086c.
376373.177:0008:0009:fixme:d3dcompiler:compile_shader Compilation target "fx_2_0" not yet supported
376373.177:0008:0009:fixme:d3dx:d3dx9_effect_init Failed to parse effect, hr 0x8876086c.
```

This can be fixed by using the redist, but yeah... I have my own parser but not my own compiler!

[...]

PR #2107, opened 2018-12-18
Add PROTON_NO_D3D10

flibitijibibo:
This acts almost like NO_D3D11, but in this case we disable not only d3d10, but also dxgi. This allows games like #2058 to work, since disabling d3d10 alone just makes the application confused when it can load dxgi but not d3d10.


tingvarsson:
As a side note on that, shouldn NO_D3D11 remove d3d10 as well (since we are talking dxvk and its d3d10 depends on d3d11 right?)

And not sure if needed but for NO_D3D11 there is a check if dxgi is overriden and only then is it cleared. Anyway it would be nice to align so the two paths remove dxgi overrides the same way for consistency.

flibitijibibo:
Found in 3.16-7!

PR #2192, opened 2019-01-04
Fix stdcall parameter size computation

mrexodia:
The size of each parameter has to be rounded up to a multiple of 4 (register size) before adding them together.

### DIFF
--- a/lsteamclient/gen_wrapper.py
+++ b/lsteamclient/gen_wrapper.py
@@ -10,6 +10,7 @@ import sys
 import clang.cindex
 import os
 import re
+import math
 
 sdk_versions = [
     "142",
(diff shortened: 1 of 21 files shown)


nsivov:
This looks correct in principle I think, it would be nicer if it wasn't using floating pointing truncation though. But argument sizes should only be important if you're building with MSVC or cross compile mingw-w64, is that your use case?

mrexodia:
I tried writing it with some helper function to round up to the nearest 4, but in my opinion the `math.ceil` makes it much clearer what's going on. In my case I encountered this issue when compiling with mingw-w64 (32 bit target) on Windows (without wine, the rest of the code is also pretty modified). It's just an issue I noticed and I wanted to contribute back upstream.

nsivov:
I think it's valid, and should be fixed, thanks for the patch.

nsivov:
This was merged as Valvesoftware/Proton@01c0124.

Issue #326, opened 2018-08-23
Battlerite (504370)


MuhammedKpln:
Character models are missing, otherwise everything works perfectly..

Deepin 15.7
Proton 3.7
Nvidia 390.67

System information from Steam:
https://gist.github.com/MuhammedKpln/b219c1b2fb76b5219c1908bd902918e4

kisak-valve:
Hello @MuhammedKpln, please copy your system information from steam (```Steam``` -> ```Help``` -> ```System Information```) and put it in a [gist](https://gist.github.com/), then include a link to the gist in this issue report.

MuhammedKpln:
@kisak-valve added system information as a gist.

ghost:
Same happens with Outcast: Second Contact. All character models are missing.
Nvidia 396.45
Specs: https://gist.github.com/byte1024/ae88bc29f9531b86c7b4d9187572dc27

kisak-valve:
Battlerite doesn't render character models in DX11 Mode.

Issue transferred from https://github.com/ValveSoftware/Proton/issues/664.
@Y4shin posted on 2018-08-26T12:45:19:

Technically not a supported game, but works pretty well except for that issue. Launching in DX9 mode fixes it.

PC Info: https://gist.github.com/Y4shin/3cd43dded169c4ecc738cf6899c68eae
Issue consistent across Proton 3.7-3 and 3.7-4
***
@doitsujin commented on 2018-08-26T16:14:37

It's a Unity Engine game and likely requires Stream Output. Will be fixed in DXVK as soon as an extension is publicly available.

kisak-valve:
Battlerite crashes while connecting to a match after the recent update

Issue transferred from https://github.com/ValveSoftware/Proton/issues/1524.
@ogdenwebb posted on 2018-09-20T22:28:58:

# Compatibility Report
- Name of the game: Battlerite
- Steam AppID of the game:  504370

## System Information
- GPU: GTX 650 TI Boost
- Driver/LLVM version:  nvidia 396.54 
- Kernel version: 4.18-5
- Link to full system information: [here](https://gist.github.com/ogdenwebb/e98cd9e7f76e741b2f7f894598e1628d)
- Proton version: 3.7.6

## Symptoms 
Today after the recent update Battlerite(i.e. 1.8.0) I can't play a match. Menu is still working well, but when I start a match or a practice the game will almost immediately crash while connecting.

[...]

PR #2246, opened 2019-01-20
[#326] Adds ISteamNetworkingSocketsSerialized interfaces

xomachine:
Workaround for https://github.com/ValveSoftware/Proton/issues/326

I don't know in what sdk versions the `ISteamNetworkingSocketsSerialized` interface is declared, so here are declarations for 002 and 003 interface versions put into 1.39 and 1.40 sdk folders. May be somebody who has access to the actual headers can put declarations to the right place.


aeikum:
Thanks, we'll take a look.

aeikum:
@xomachine I added support for SteamNetworkingSocketsSerialized002 in 3.16-7... and then Battlerite updated to use SteamNetworkingSocketsSerialized003 on the day before we shipped. So we'll need to update again for that. Thanks for reporting and submitting the fix. I'll try to have support for v3 in the next release.

PR #2291, opened 2019-01-31
When copying ignore EPERM error

davispuh:
When `$PREFIX/drive_c` points to location which is owned by another user (eg. NTFS drive mounted with different `uid=,gid=,fmask=`) then `shutil.copy` will fail with `Operation not permitted` because after successfully copying file it will try to `chmod` which won't work but permissions might be already correct to successfully read/write/execute due to group and ACLs.

```
  File "~/.local/share/Steam/steamapps/common/Proton 3.16 Beta/proton", line 301, in <module>
    shutil.copy(steamdir + "/legacycompat/" + f, dstfile)
  File "/usr/lib/python3.7/shutil.py", line 246, in copy
    copymode(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.7/shutil.py", line 144, in copymode
    chmod_func(dst, stat.S_IMODE(st.st_mode))
PermissionError: [Errno 1] Operation not permitted: '~/.local/share/Steam/steamapps/compatdata/221380/pfx//drive_c/Program Files (x86)/Steam/steamclient.dll'
```

This PR, fixes it by just ignoring this case. In case we actually don't have correct permissions copied we'll still see it later anyway when those needed files are actually accessed.

### DIFF
--- a/proton
+++ b/proton
@@ -8,6 +8,7 @@ import filecmp
 import json
 import os
 import shutil
+import errno
 import struct
 import subprocess
 import sys
@@ -78,7 +79,7 @@ def upgrade_pfx(old_ver):
             log("Removing newer prefix")
             if old_proton_ver == "3.7" and not os.path.exists(os.environ["STEAM_COMPAT_DATA_PATH"] + "/tracked_files"):
                 #proton 3.7 did not generate tracked_files, so copy it into place first
-                shutil.copy(basedir + "/proton_3.7_tracked_files", os.environ["STEAM_COMPAT_DATA_PATH"] + "/tracked_files")
+                try_copy(basedir + "/proton_3.7_tracked_files", os.environ["STEAM_COMPAT_DATA_PATH"] + "/tracked_files")
             remove_tracked_files(os.environ["STEAM_COMPAT_DATA_PATH"])
             return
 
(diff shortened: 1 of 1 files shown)


aeikum:
This makes me nervous. If this is broken, what else is not going to work? Pushing the failure further down the line seems like the wrong solution.

davispuh:
> This makes me nervous. If this is broken, what else is not going to work? Pushing the failure further down the line seems like the wrong solution.

It's not really pushing it down because this issue is with Proton and usage of `chmod`. It's unlikely that started application/game will try to change file's permissions using `chmod` while also checking result of that call and besides Wine could be patched to always return success.
I tested that this PR lets me run "Alien Shooter" with `drive_c` which is owned by another user but it's group's permission allows reading/writing and thus this allows to share same folder across multiple users.

Anyway I ended up doing something else - using [bindfs](https://bindfs.org/) I mounted directory owned by another user to different place with `mirror` option which will show it as owned by current user.

in my fstab
```
/mnt/SteamReal  /mnt/Steam            fuse.bindfs    noauto,mirror=@users,create-as-user,create-for-group=users,x-systemd.automount        0 0
```

When viewed by user1
```
$ ls -la /mnt/
total 68
drwxr-xr-x 1 root   root        204 feb  2 17:09 ./
drwxr-xr-x 1 root   root        122 jan 14 19:28 ../
drwxrwxr-x 1 user1  users 12288 feb  5 13:44 Steam/
drwxrwxr-x 1 steam  users 12288 feb  5 13:44 SteamReal/
``` 

and by user2
```
$ ls -la /mnt/
total 68
drwxr-xr-x 1 root   root        204 feb  2 17:09 ./
drwxr-xr-x 1 root   root        122 jan 14 19:28 ../
drwxrwxr-x 1 user2  users 12288 feb  5 13:44 Steam/
drwxrwxr-x 1 steam  users 12288 feb  5 13:44 SteamReal/
``` 

This makes `chmod` succeed and allows to share all downloaded games by all users without need to copy/duplicate them.
Only downside is that it affects disk/IO performance a lot and is quite slower.

kisak-valve:
Merged https://github.com/ValveSoftware/Proton/commit/817828ec72acad36d9cebc00f5703b17166aa22d.

PR #2371, opened 2019-03-01
vrclient: Translate paths for SetActionManifestPath

hdmap:
Games calling SetActionManifestPath will currently get back an error and have no input because the linux vr runtime is receiving an untranslated windows/wine path. This affects all games that use the newer steamvr input system.

Most of this is copied over from lsteamclient.

I tested this with/for COMPOUND (steamappid 615120, and the demo is an older build that doesn't use the new input system so it's not affected) but a free game showing this is Sopwith VR (steamappid 998660).


aeikum:
Thanks a lot for working on this, we'll take a look soon.

aeikum:
This was merged in 4.2-1 as 1ff16641fd350682c30420dada6f92a3e6bce421. Thanks again!

PR #2503, opened 2019-04-04
Update CURRENT_PREFIX_VERSION to latest

Zlopez:
I noticed that this constant wasn't changed with latest release, so here is a patch for it.

### DIFF
--- a/proton
+++ b/proton
@@ -19,7 +19,7 @@ from filelock import FileLock
 #To enable debug logging, copy "user_settings.sample.py" to "user_settings.py"
 #and edit it if needed.
 
-CURRENT_PREFIX_VERSION="4.2-1"
+CURRENT_PREFIX_VERSION="4.2-2"
 
 PFX="Proton: "
 ld_path_var = "LD_LIBRARY_PATH"


Plagman:
I think we only need to change this when necessary, eg. if the prefix structure or contents changes in incompatible ways.

Zlopez:
Strange, the games are still using the 4.2-1, till I change this manually. I noticed the same behavior on 3.16-8, when the games still used 3.16-1.

Zlopez:
It looks like I was wrong, I tried one game today with `PROTON_LOG=1` and the version used was `4.2-2`.

I was just confused that I didn't saw this message when I started the game after update from `4.2-1` to `4.2-2` `log("Upgrading prefix from " + str(old_ver) + " to " + CURRENT_PREFIX_VERSION + " (" + os.environ["STEAM_COMPAT_DATA_PATH"] + ")")`

aeikum:
Yeah @Plagman is right. This is the "prefix version" which is separate from the "Proton version." We bump it when we need to do something special when changing to a new Proton version.

PR #2529, opened 2019-04-12
proton: Place DXVK logs into the user home folder.

jp7677:
This will prevent DXVK logs all over the steamapps directory, but places them in the same folder as the wine logs.

### DIFF
--- a/proton
+++ b/proton
@@ -278,6 +278,9 @@ if "SteamGameId" in env:
 else:
     env["WINEDEBUG"] = "-all"
 
+if env["DXVK_LOG_LEVEL"] != "none":
+    env["DXVK_LOG_PATH"] = os.environ["HOME"]
+
 def create_fonts_symlinks(prefix_path):
     fontsmap = [
         ( "LiberationSans-Regular.ttf", "arial.ttf" ),


aeikum:
I'd rather have this log output dumped to the same file as the rest of the Proton logging. In other words, is there a way to configure DXVK to output its logging to stdout or stderr?

jp7677:
According to https://github.com/doitsujin/dxvk/blob/master/src/util/log/log.cpp#L47 DXVK writes its log statements simultaneously both to file and to stderr.

aeikum:
OK, so we should have what I want already. If it's not appearing in the Proton log, then probably the game is redirecting stderr someplace else (I know Unity does this a lot). That's a separate issue, which should be solved if we can ever build DXVK as a native Linux library.

I'm resistant to the idea of cluttering up the user's home folder with even more log files. If it's useful for you, you can add it to your user_settings.py to avoid having it overwritten by future Proton updates.

jp7677:
> I'm resistant to the idea of cluttering up the user's home folder with even more log files. If it's useful for you, you can add it to your user_settings.py to avoid having it overwritten by future Proton updates.

Enabling DXVK logs is a deliberate option in Proton, I guess people that enable that option want to have easy access to the logs. I get your point with more cluttering and agree that having just one log file would be the best solution, though for now having the logs in your home folder would be, in my opinion, more convenient than searching within the folder structure of the game directory.

That said placing that option into the user settings is indeed a sensible option which certainly works for me, even better than setting that env variable in the game launch options which I did until now. Should I keep this PR open or do you prefer to close it?

aeikum:
One other complication (which you couldn't know about) is I have an upcoming commit that enables DXVK logging when `PROTON_LOG` is set, to more closely match the default user_settings.py settings. That would then trigger your extra log file, too.

Anyway yeah, it's a reasonable suggestion, but I think just putting it in user_settings seems like a good compromise.

jp7677:
Thanks a lot for your response and I'm looking forward to upcoming releases!
(Not just for the logging improvements but also in general :), you and your colleagues are doing very cool work here!)

aeikum:
Thanks, glad you are enjoying it. Stay tuned, exciting things are coming :)

PR #2530, opened 2019-04-12
Use --recurse-submodules with git-clone

leifmetcalf:
This way you only have to copy-paste one line.


aeikum:
Cool, didn't know about that option. Will merge.

aeikum:
This was merged as ecb0f0bfe70b49b6523a5240ca055f0e6c9d96e1.  Thanks!

PR #2590, opened 2019-04-23
makefile: Move generated files into build directory

zzhiyi:
@aeikum 

This way, we can run make again without rerunning configure.sh
after a clean rsync, which should save several minutes for every build.

Signed-off-by: Zhiyi Zhang <zzhang@codeweavers.com>


aeikum:
Looks fine to me. Will merge.

Issue #818, opened 2018-08-28
Commandos 2 Men of Courage (6830)


Zero86Sk:
This can easily go onto the whitelist. Works as intended.

My Sytem info:
https://gist.github.com/Zero86Sk/d629f657c1ae861c36657cdf238a7765

AwesamLinux:
Distro: Linux Mint 19.1 Tessa (64 bit)
GPU: Vega 56
Driver/LLVM version: Mesa 19.1.0-devel/9.0.0
Kernel version: 5.1.1-050101-lowlatency
Proton version: 4.2-3

Game is playable, but cut-scene videos are skipped. Also the game launches behind the desktops taskbar like this:
![Screenshot from 2019-05-12 19-59-02](https://user-images.githubusercontent.com/43387963/57586165-2ef18a80-74f2-11e9-9b97-29711fac0414.jpg)

Need to alt+tab between desktop and the game to fix it.

[media_info.txt](https://github.com/ValveSoftware/Proton/files/3170581/media_info.txt)
[steam-6830.log](https://github.com/ValveSoftware/Proton/files/3170582/steam-6830.log)
[system_info.txt](https://github.com/ValveSoftware/Proton/files/3170583/system_info.txt)

In the log I see
`19994.801:0027:0028:fixme:quartz:AsyncReader_QueryInterface No interface for {2dd74950-a890-11d1-abe8-00a0c905f375}!`

#1464 The game uses quartz and the videos (stored as .pop files) are: WMV, VC-1 + WMA

Steap:
I disagree: the game randomly crashes early in the first tutorial mission.

[My system info](https://gist.github.com/Steap/175e51773a0cca16beb8ef7f40e99166)

[The crash log](https://gist.github.com/Steap/2186b5bd960ef8ae84c7cfa254a97e10)

ranplayer:
So I've tested this game with Proton 9.0-4, 10.0-2 (RC) and Experimental (2025-07-11).  My findings:
- 9.0-4: 
  - it seems to work well (just the videos are skipped)
- 10.0: 
  - videos work, but no sound
  - soft mouse locks during gameplay (sometimes even on the menu)

Proton 10 logs: [steam-6830.tar.gz](https://github.com/user-attachments/files/21217796/steam-6830.tar.gz)

Specs:
- OS: Arch
- Kernel: 6.12.29
- GPU: RX 6700 XT (Mesa 25.0-5)
[...]

Issue #816, opened 2018-08-28
Commandos Beyond the Call of Duty (6810)


Zero86Sk:
Works as Intended on my system.

[My System](https://gist.github.com/Zero86Sk/d629f657c1ae861c36657cdf238a7765)

andrebadaia:
Confirmed.

Distro: Linux Mint 19 Tara
Processor: AMD Ryzen 7 1700
GPU: GTX 1070, Nvidia proprietary driver 396.54
RAM: 32GB
Kernel (custom) - 4.17.14 (optimized for Ryzen)
Resolution: 1920x1080 (In game 1280x720)
Proton 3.7-3

Steap:
I started playing this game, and it works well for the most part. Only two issues so far:

1) The videos cannot play, just like in [Commandos : Behind Enemy Lines](https://github.com/ValveSoftware/Proton/issues/176). I think it is again related to bug #1464.

2) Harder to reproduce: sometimes loading a savegame crashes the game. After restarting the game, it is possible to load the same savegame. Here is the [log](https://gist.github.com/Steap/91a8605ffe35b8e2d637044f03edfe71)


Here is my [system info](https://gist.github.com/Steap/f16088fa95aa0054db4c887fb19db524).

AwesamLinux:
Distro: Linux Mint 19.1 Tessa (64 bit)
GPU: Vega 56
Driver/LLVM version: Mesa 19.1.0-devel/9.0.0
Kernel version: 5.1.1-050101-lowlatency
Proton version: 4.2-3

I can confirm that the game is playable but the videos do not work.

[media_info.txt](https://github.com/ValveSoftware/Proton/files/3170369/media_info.txt)
[steam-6810.log](https://github.com/ValveSoftware/Proton/files/3170370/steam-6810.log)
[system_info.txt](https://github.com/ValveSoftware/Proton/files/3170371/system_info.txt)

 in the log i see:
`18553.621:0027:0028:err:quartz:GetClassMediaFile Media class not found`

#1464 The game use quartz and the videos are: AVI, Cinepak + PCM

PR #2678, opened 2019-05-12
Add request for Proton log to Whitelist Request template

HonkingGoose:
### What changes?
There's an extra item on the list asking for a Proton log.

### Rationale for change:
Some games are reported as whitelist-ready while not actually being free of issues (cutscenes/cinematics, other bugs). By asking for a log we can check if the whitelisted game is actually free of issues.

Example: #816 and #818 were initially reported as whitelist-ready but later turn out to not be whitelist-ready due to issues with video playback.

### Benefits of this change:
- Filtering out wrongly filed whitelist requests by checking the log for errors.
- Whitelist request tag will only have proper whitelist ready games.
- If whitelist request was "proper" we now have a known good log in case of regression.

### DIFF
--- a/.github/ISSUE_TEMPLATE/whitelist-request.md
+++ b/.github/ISSUE_TEMPLATE/whitelist-request.md
@@ -37,4 +37,6 @@ about: Games tested and found to have no issues.
    this issue to prevent chaos by too much info in one place.
 5. Please search for open issues and pull requests by the name of the game and
    find out whether they are relevant and should be referenced above.
+6. Please add `PROTON_LOG=1 %command%` to the game's launch options and drag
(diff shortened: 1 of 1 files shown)


aeikum:
Thanks for the suggestion! I ran this by our whitelisting team and they said it wouldn't be terribly useful for their purposes. I think asking users to post a log could present a barrier that would cause some people to give up on their report. If it's not useful for our whitelisting team, I think it makes more sense not to have this as a requirement.

HonkingGoose:
Hi @aeikum, thanks for looking at my PR, and considering if it adds value.

If the team dealing with the whitelists says: "A Proton log would not be helpful, and only raise unnecessary barriers." then I think it makes sense to listen to them. :smile: 

We'll keep things as-is then, I'll go close the PR.

PR #2707, opened 2019-05-18
Fix for Nelke & The Legendary Alchemists

ben-romer:
Hello,

This is just a trivial fix that gets Nelke & the Legendary Alchemists working correctly, and it might fix minor problems in other titles if they have the same problem. If you try to save in Nelke, it blows up with an error in CreateDirectory() and refuses to save because the save directory already exists. It does that because of a fixme that returns the wrong error code.

Built and Tested with the latest version of Steam and it worked perfectly for me.


kisak-valve:
~Hello @ben-romer, this patch should also be sent to upstream wine if it hasn't been already.~

Possibly fixed upstream at https://source.winehq.org/git/wine.git/commit/9b6d198a3c7c8a02aa69c0d3d11829712e9778a6.

ben-romer:
Ah yes, that upstream patch looks like it would work too. I'll test it out with Nelke and close my pull request if it does. Thanks!

aeikum:
Cool, thanks for letting us know. I can pick this back to Proton 4.2. To remind me to do that, feel free to leave this PR open or open a new one in the Wine submodule.

ben-romer:
I'll open a new one in the wine submodule with the upstream patch instead - it works. :)

Issue #2742, opened 2019-05-26
steam-runtime bison package too old to compile wine 4.7+


GloriousEggroll:
As I regularly make custom builds of proton, I came across this today while attempting to compile 4.9:

`configure: error: Your bison version is too old. Please install bison version 3.0 or newer.
../proton/build/makefile_base.mak:755: recipe for target 'obj-wine32/Makefile' failed
make[1]: *** [obj-wine32/Makefile] Error 1
`
So this was apparently a change between wine 4.6 and 4.7:

https://github.com/wine-mirror/wine/commit/a4c93936c9493f7619e1877b522eeb809a390dfe

And when we check the steam runtime version:
```
vagrant@debian9:~/proton$ docker run steam-proton-dev bash -c "bison --version"
bison (GNU Bison) 2.5

vagrant@debian9:~/proton$ docker run steam-proton-dev32 bash -c "bison --version"
bison (GNU Bison) 2.5
```
Further investigation after reverting the commit mentioned above also shows this error when compiling:
```
../../../../proton/wine/tools/widl/parser.y:321.9-19: %define variable `parse.error' is not used
tools/winebuild/winebuild -w -o dlls/netapi32/libnetapi32.def --def -m32 --export \
  ../../proton/wine/dlls/netapi32/netapi32.spec
make[1]: *** [Makefile:395: parser.tab.h] Error 1
```
and later also:
```
../../../../proton/wine/dlls/d3dcompiler_38/../d3dcompiler_43/hlsl.y:906.9-19: %define variable `parse.error' is not used
make[1]: *** [Makefile:541: hlsl.tab.h] Error 1
```
So... I figured already there may be more files affected by this change:

```
$ grep -rnw . -e "define parse.error verbose"

./dlls/wbemprox/wql.y:190:%define parse.error verbose
```

Caused by these commits:
https://github.com/wine-mirror/wine/commit/5ea4d5971bff3430c29f1da31166b0a2565347ab
https://github.com/wine-mirror/wine/commit/2d27c13d5ea76ec80e2a67272c8eafe8527e6af0
https://github.com/wine-mirror/wine/commit/6d4cdeb658e1cfd7c73643f674a0043006970b44

So for future proofing, you'll either have to revert the commits, or update the steam runtime, or add bison to the build process.

Just lookin' out for future proofing.

aeikum:
[...]

PR #2754, opened 2019-05-30
add bison 3.0.4 to proton build

GloriousEggroll:
solves https://github.com/ValveSoftware/Proton/issues/2742

### DIFF
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -231,6 +231,13 @@ CMAKE_OBJ64 := ./obj-cmake64
 CMAKE_BIN32 := $(CMAKE_OBJ32)/built/bin/cmake
 CMAKE_BIN64 := $(CMAKE_OBJ64)/built/bin/cmake
 
+BISON := $(SRCDIR)/bison
+BISON_OBJ32 := ./obj-bison32
+BISON_OBJ64 := ./obj-bison64
+BISON_BIN32 := $(BISON_OBJ32)/built/bin/bison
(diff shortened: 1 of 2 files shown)


aeikum:
Thanks! I didn't review the Makefile changes in depth, but we definitely want to be using a submodule, not importing the source directly. http://savannah.gnu.org/git/?group=bison

GloriousEggroll:
Not a problem, corrected the PR, added 3.0.4 as a submodule. Don't mind the name change, I added the new commits from my work machine.

GloriousEggroll:
I've noticed when using bison from git, you must run ./bootstrap before configuring, also even after bootstrapping I still ran into some errors I wasn't able to troubleshoot that were not present in the tarball. Tarballed version worked fine :/

aeikum:
Thanks. I'll probably let this sit until we begin work on the next Wine upgrade, and review it then.

aeikum:
Hey @GloriousEggroll, I ended up going back to using a tarball, because generating those files is pretty tough. Instead of adding the files to repo, I just download the tarball at build-time. You can check out the changes I made here: https://github.com/aeikum/Proton/tree/proton_4.2-bison

I left your authorship on the commit, since the bulk of the work is yours. If you'd like me to change it in any way, let me know.

aeikum:
This was added for the 4.11 build as d7819c7bc5489b4e30c3b0782ccde78c5cd41cd8. Thanks!

PR #2852, opened 2019-07-03
Added Proton variable usage example to readme

echaskaris:


aeikum:
I the preceding sentence not clear enough? I'm not opposed to making this more clear, but I think we should re-word that sentence instead of adding another.

echaskaris:
I don't think it's clear enough, I substituted 'command' with the proton variables and left PROTON_VARIABLE=1 as is. As it turns out, that's not correct.

aeikum:
I changed the wording in 988271e7c2df75e66bbdcb7cbdcf8e3be5ff1159. Do you find this new wording acceptable?

echaskaris:
Yes

aeikum:
Great. Thanks for the suggestion!

PR #2893, opened 2019-07-16
Include a vagrant executible in the source

pmorgan3:
The current rubygem way of installing vagrant (that's in the readme) will just lead you to installing this executable anyway. Might as well include it with the source to make installation easier.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -73,48 +73,48 @@ help:
 	@echo "      in the \"mytest\" build directory and place it into vagrant_share/dsound/."
 
 vagrant:
-	vagrant up
-	vagrant rsync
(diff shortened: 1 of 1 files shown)


dreamer:
40MB binary file in git repo without using `git-lfs`? This is a bad idea - it will make the repository grow by 40MB each time someone will commit newer version of this binary. Also, you forgot to include license.

pmorgan3:
> 40MB binary file in git repo without using `git-lfs`? This is a bad idea - it will make the repository grow by 40MB each time someone will commit newer version of this binary. Also, you forgot to include license.

Good point! You're right it was a bad idea

PR #2971, opened 2019-08-10
proton: Refactor dump_dbg_scripts()

pchome:
I merged this code using an plain text editor, just to figure out for myself if there any difference in the debug script generators. Resulting code wasn't even tested.

Just in case, if it could be useful.

Why some tricks used:
* single quotes to avoid multiple `\"`, a few rare `\'` was added instead
* several separate text strings on new line are treated by `python` as single line text string
  _was used to made the code similar (by indentation) to the bash script it supposed to generate_
  _(with the variable success, obviously, after everything was collapsed)_
* `for idx, val in enumerate(list):` was used to drop `first = True` variable,
  but such loops was replaced by `'arr=("' + '" "'.join(list) + '")'` construction late, to generate `bash` array in a single line

### DIFF
--- a/proton
+++ b/proton
@@ -546,27 +546,32 @@ class Session:
         else:
             self.env["WINEDLLOVERRIDES"] = s
 
-    def dump_dbg_env(self, f):
-        f.write("PATH=\"" + self.env["PATH"] + "\" \\\n")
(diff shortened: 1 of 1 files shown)


aeikum:
The environment stuff is a good change. The rest of it seems more convoluted than it's worth.

PR #3116, opened 2019-10-02
Update hex_float.h

0x0J:


kisak-valve:
Hello @0x0J, changes to glslang should be discussed in the upstream project for a lasting effect to happen. https://github.com/KhronosGroup/glslang/blob/master/SPIRV/hex_float.h#L856 looks like the upstream code.

Closing because this is a no-op change in a third party component.

PR #3237, opened 2019-11-20
build: Move environment variables to the end of configure script.

zzhiyi:
@aeikum 

"ENV=VAR ./configure" is different from "./configure ENV=VAR".
The latter caches command line variables in config.status.
So "config.status --recheck" succeeds executing configure script
Otherwise, "config.status --recheck" often fails when only source
is updated because some variables are not cached. Examples include
being not able to find bison or vkd3d.h when they are obviously present,
which are because of missing BISON and VKD3D_CFLAGS variables.

Another way to fix the build issue is to mark those variables needed to be kept as [precious variables](https://www.gnu.org/software/autoconf/manual/autoconf-2.69/html_node/Setting-Output-Variables.html). But then we have to keep track of those variables as well.

Related links:
1. https://stackoverflow.com/questions/13848154/passing-environment-variables-to-autoconfs-configure
2. https://unix.stackexchange.com/questions/290381/why-does-configure-take-variables-as-arguments

### DIFF
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -834,33 +834,36 @@ WINE32_MAKE_ARGS := \
 $(WINE_CONFIGURE_FILES64): SHELL = $(CONTAINER_SHELL64)
 $(WINE_CONFIGURE_FILES64): $(MAKEFILE_DEP) | faudio64 vkd3d64 $(WINE_OBJ64) bison64
 	cd $(dir $@) && \
-		STRIP=$(STRIP_QUOTED) \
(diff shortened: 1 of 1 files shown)


aeikum:
Great, thanks. This looks like the right way to fix what faf2b8271ab505c38b8492228727c8896b1620a4 was trying to fix. Will merge.

aeikum:
This was merged as aee1b47925de6d06b4870b8f60675ba13e7ec269

aeikum:
This was merged as aee1b47925de6d06b4870b8f60675ba13e7ec269

PR #3240, opened 2019-11-20
RFC: Disable esync when running the install script.

Guy1524:
Origin fails to install/update when esync is enabled.  We can work around this by disabling esync during the install script.  If I'm not mistaken, wineserver is restarted before the game itself is run, so this should be safe.  One concern though is how this would affect non-steam games run through Proton.

### DIFF
--- a/proton
+++ b/proton
@@ -459,10 +459,10 @@ class Session:
         self.check_environment("PROTON_OLD_GL_STRING", "oldglstr")
 
         if not "noesync" in self.compat_config:
-            self.env["WINEESYNC"] = "1"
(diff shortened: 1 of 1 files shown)


zzhiyi:
I don't know enough about esync. If it's esync's fault that Origin don't install, I think esync should be fixed . Or does it trigger a behavior that esync can't handle by design?

Leopard1907:
@zzhiyi

https://bugs.winehq.org/show_bug.cgi?id=48089#c4


Yes , esync needs that to be implemented but there is no eta.

PR #3404, opened 2020-01-03
Update makedirs function to not supress OSErrors

SkyLeite:
The current implementation of makedirs suppresses every OSError possible and assumes the directory already exists when one is raised. This is problematic as there are many reasons why the error could occur, and suppressing it could make it hard to trace the problem back in the future.

My implementation fixes this by checking whether the path exists before attempting to create it, and logs every other OSError appropriately.

I also took the liberty of changing a stray `os.makedirs` call to our `makedirs` function.

### DIFF
--- a/proton
+++ b/proton
@@ -49,8 +49,7 @@ def makedirs(path):
             return
         
         os.makedirs(path)
-    except OSError as err:
-        log("Error when creating directory {0}: {1}".format(path, err))
+    except FileExistsError:
         pass
 
 def try_copy(src, dst):
@@ -170,9 +169,9 @@ class CompatData:
                 try:
                     if not any(os.scandir(d)):
                         os.rmdir(d)
-                except OSError as err:
-                    log("Error deleting directory {0}: {1}".format(d, err))
-                    pass
+                except OSError as e:
+                    if e.errno != errno.ENOTEMPTY:
+                        raise
 
         os.remove(self.tracked_files_file)
         os.remove(self.version_file)
--- a/proton
+++ b/proton
@@ -45,9 +45,11 @@ def file_is_wine_fake_dll(path):
 
 def makedirs(path):
     try:
+        if os.path.lexists(path):
+            return
+        
         os.makedirs(path)
     except OSError:
-        #already exists
         pass
 
 def try_copy(src, dst):
--- a/proton
+++ b/proton
@@ -168,7 +168,7 @@ class CompatData:
                         dirs.append(path)
             for d in dirs:
                 try:
-                    if len(os.listdir(d)) != 0:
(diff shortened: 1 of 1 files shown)


aeikum:
Looks reasonable, thanks. Will merge.

SkyLeite:
Updated a different occurrence of this as well. We're also ignoring errors in the `set_dir_casefold_bit` but I don't know what that function is supposed to do so I'm not touching it for now.

aeikum:
The logic seems fine, but is `listdir` and `len` the best way to determine if a directory is empty? This code isn't terribly perf-critical, but maybe doing something with a `scandir` iterator would be more efficient? (I'm not a Python expert; maybe the optimizer is smarter than I think?)

SkyLeite:
From what I could gather from [the scandir proposal](https://www.python.org/dev/peps/pep-0471/), that seems to be correct. The latest commit should address that.

proton
```diff
@@ -165,9 +168,10 @@ class CompatData:
                         dirs.append(path)
             for d in dirs:
                 try:
-                    os.rmdir(d)
-                except OSError:
-                    #not empty
+                    if not any(os.scandir(d)):
```
nanonyme: This introduces a race condition. Please rather change OSError to FileNotFoundError since this is Python 3

proton
```diff
@@ -45,9 +45,12 @@ def file_is_wine_fake_dll(path):
 
 def makedirs(path):
     try:
+        if os.path.lexists(path):
+            return
+        
         os.makedirs(path)
-    except OSError:
-        #already exists
+    except OSError as err:
```
nanonyme: Use FileExistsError instead of OSError

proton
```diff
@@ -165,9 +168,10 @@ class CompatData:
                         dirs.append(path)
             for d in dirs:
                 try:
-                    os.rmdir(d)
-                except OSError:
-                    #not empty
+                    if not any(os.scandir(d)):
+                        os.rmdir(d)
+                except OSError as err:
```
nanonyme: Instead
```
except OSError as e:
    if e.errno != errno.ENOTEMPTY:
      raise
```

SkyLeite: As far as I could gather checking if the directory is empty before attempting to do the operation is less expensive than letting it throw an exception and catching it. Would we still prefer going with the latter approach?

nanonyme: There's no measurable performance difference between the two (although the deletion and seeing whether it fails is most likely faster as it results in less syscalls on average) but with the exception handling race conditions are impossible. That's typically recommended way to handle these things.

SkyLeite: Makes sense. I hadn't considered race conditions. I'll make the changes in a bit :)

PR #3513, opened 2020-02-10
proton: Add option to disable DXVK for d3d9.

redmcg:
Whilst the main game of Elite Dangerous (the "client") uses D3D11, the launcher uses D3D9 (via WPF in dotnet40).

Unfortunately, DXVK doesn't appear to render the launcher correctly. For example, the 'PLAY' button (which launches the game) does not appear.

By disabling just D3D9 for DXVK, the launcher will render correctly but the main game will still run using DXVK.

This patch introduces the option 'PROTON_NO_D3D9', which allows the disabling of just D3D9 rendering via DXVK.

Here's a couple of screen shots to better highlight the Elite Dangerous issue:
With DXVK rendering D3D9
image

After applying the suggested patch and using the following launch options:
image

image

### DIFF
--- a/proton
+++ b/proton
@@ -449,6 +449,7 @@ class Session:
             self.check_environment("PROTON_USE_WINED3D11", "wined3d")
         self.check_environment("PROTON_NO_D3D11", "nod3d11")
         self.check_environment("PROTON_NO_D3D10", "nod3d10")
+        self.check_environment("PROTON_NO_D3D9", "nod3d9")
         self.check_environment("PROTON_NO_ESYNC", "noesync")
         self.check_environment("PROTON_NO_FSYNC", "nofsync")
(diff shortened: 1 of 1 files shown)


aeikum:
I don't think this option has a lot of use. I think it would be better to report this to the DXVK devs, who can hopefully fix it quickly, so we won't need this option going forward. In the meantime, you can continue to use 4.11.

redmcg:
Understood. I'll raise this with the DXVK devs.

doitsujin:
We're aware of the issue but have not yet found a way to actually debug it.

redmcg:
Thanks @doitsujin. I found the open issue:
https://github.com/doitsujin/dxvk/issues/1408

misyltoad:
The underlying issue in DXVK should be fixed now.

PR #3518, opened 2020-02-12
Use ShellExecute when not launching .exes

misyltoad:
This behaviour more closely matches what Steam on Windows does when
launching things that aren't executables.

I tested this with some wacky edge-cases on my own appid (launching image files, random blah://frog links etc.) both on Windows and via. Proton this appears to mirror the behaviour fairly well.

### DIFF
--- a/steam_helper/steam.cpp
+++ b/steam_helper/steam.cpp
@@ -142,7 +142,45 @@ static BOOL WINAPI console_ctrl_handler(DWORD dwCtrlType)
     return TRUE;
 }
 
-static HANDLE run_process(void)
+static BOOL should_use_shell_execute(WCHAR* cmdline)
+{
+    BOOL use_shell_execute = TRUE;
+
+    BOOL quoted = FALSE;
+    WCHAR* executable_name_end = cmdline;
+    
+    /* find the end of the first arg...*/
+    while (*executable_name_end != '\0' &&
+           (*executable_name_end != ' ' || quoted) &&
+           (*executable_name_end != '"' || !quoted))
+    {
+        quoted ^= *executable_name_end == '"';
+
+        executable_name_end++;
+    }
+
+    /* backtrack to before the end of the arg
+     * and check if we end in .exe or not
+     * and determine whether to use ShellExecute
+     * based on that */
+    executable_name_end -= strlen(".exe");
+
+    if (executable_name_end >= cmdline)
+    {
+        static const WCHAR exeW[] = {'.','e','x','e',0};
+
+        WINE_TRACE("Command ends in %s, (quoted: %s) \n",
+            wine_dbgstr_w(executable_name_end),
+            wine_dbgstr_a(quoted ? "yes" : "no"));
+
+        if (!wcsncmp(executable_name_end, exeW, wcslen(exeW)))
+            use_shell_execute = FALSE;
+    }
+
+    return use_shell_execute;
+}
+
+static HANDLE run_process(BOOL* should_await)
 {
     WCHAR *cmdline = GetCommandLineW();
     STARTUPINFOW si = { sizeof(si) };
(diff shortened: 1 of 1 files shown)


GloriousEggroll:
Just some feedback on this - Doom II and Ultimate Doom both launch, clustertruck launches.  First launch progress bars work. zach-like does not launch still.

note: zach-like fails even on normal proton. it opens a small black console box then closes.

misyltoad:
This patch is not meant to address the Zach-Like issue, that is entirely separate.

I will have a look into fixing that tomorrow :smile: :frog:

GloriousEggroll:
zach-like was brought up in the previous bug, that's the only reason I brought it up. seems it has a different issue.

aeikum:
@Joshua-Ashton Thanks. Instead of trying to parse the command, can we use the result from SHGetFileInfo to determine whether to use ShellExecute or CreateProcess? E.g. if non-zero (i.e. a PE file or COM file), use CreateProcess; else, use ShellExecute. I tested this out with Origin and Jedi Fallen Order and it seems to work fine.

For non-game processes, is it important that Steam.exe stick around? Can we just exit immediately in the ShellExecute case instead of introducing another call to __wine_make_process_system?

I can make all these changes, just want to run them by you. If they sound good I'll share my changes with you for review.

aeikum:
@Joshua-Ashton Here's something you could review: https://github.com/aeikum/Proton/commit/d43dd89d63058b337d6a2ffbf6bd3d338b26466e

misyltoad:
So I did some extra debugging of this issue:

When we call CreateProcessW, we should always pass in the flag `CREATE_UNICODE_ENVIRONMENT`.

`SHGetFileInfoW` is never called, so I don't think this is a good determiner of things. I think we should stick as close to what native Steam does as possible.

steamclient.dll does end up testing if the application name isn't NULL [do we need to add this check?] and ends in ``.exe``, like I do in my patch:
```cpp
if ( firstArg && !V_StringFindSuffix(firstArg, ".exe") )
```

If it ends in ``.exe`` then we go on the CreateProcessW path like this:

```cpp
   DWORD dwCreationFlags = SomeFunctionArgThatIsZeroWhenCalled | CREATE_UNICODE_ENVIRONMENT;
   
   // Something about starting with SUSPEND for game overlay crap.
   
    if ( CreateProcessW(
           lpApplicationName,
           lpCommandLine,
           NULL,
           NULL,
           FALSE,
           dwCreationFlags,
           lpEnvironment,
           lpCurrentDirectory,
           &StartupInfo,
           &ProcessInformation) )
       ...
```

and if it doesn't then it attempts to create `steam_appid.txt` (and if that fails, it will Assert) and then calls through to some helper which ends up calling either ShellExecuteExW via tier0_s.dll [This seems to have changed from doing ``ShellExecuteA`` with the open verb via vgui2.dll since I last looked at it... Unless I frogged something up the first time.] I will look at that.

> For non-game processes, is it important that Steam.exe stick around?

I'd imagine it would be given some titles on Steam end up calling their own proprietary launchers like ``myfroglauncher://rungame//BigFrogsLatestGreatestReturn2``, which may have some shitty DRM checks with linking that could check if Steam.exe is alive.
I don't think any that depend on that exist right now, but I am just thinking ahead :frog:

aeikum:
Sure, we can add that flag in a separate commit.

We're currently using SHGetFileInfoW to determine if the application is a console application, and pass in the `CREATE_NEW_CONSOLE` flag so the application will run in a console. That's what the ZACH-LIKE program uses. Since we're already using this function it seemed straightforward to use it for both purposes. (Also I really dislike string parsing; it's error prone and we already do a lot.)

> I'd imagine it would be given some titles on Steam end up calling their own proprietary launchers l

Steam launches the installscript stuff without the SteamGameId variable set; I call these "non-game processes". When it gets around to the game's launch command, it does have that variable set, so this is the "game process." Whether it's actually the Game.exe or not isn't important. My point was more that we don't want a bunch of Steam.exe processes hanging around. We should have just one, which is the one we start alongside the "game process".

misyltoad:
There is no logic in Steam to pass in `CREATE_NEW_CONSOLE` as a flag so the problem is elsewhere:

When I start ZACH-LIKE on Windows I get `CREATE_UNICODE_ENVIRONMENT | CREATE_SUSPENDED`. (latter for the weird game overlay injection stuff that we don't need.)

The real reason ZACH-LIKE fails is that the Steam helper exe is built for the Console subsystem (this defaults to `-mconsole`), not Windows. (`-mwindows`.)
[ie. it needs to be a GUI app for the process to spawn its own console automagically without that flag]

misyltoad:
> My point was more that we don't want a bunch of Steam.exe processes hanging around. 

My reasoning was more for titles that don't use the Steam API or interface with Steam at all expecting Steam.exe to be present after the ShellExecute (otherwise we are already dead, unless I am mistaken in that case?)

aeikum:
This was merged, somewhat modified, as 68f982bd47bd8f20249859da825a7bbcf36600bd.

PR #3541, opened 2020-02-18
Dump WINEFSYNC

DadSchoorse:
Currently only `WINEESYNC` gets dumped to the debug scripts.

`WINEFSYNC` should be dumped, too, to mirror normal behavior.


aeikum:
Good catch, thanks.

aeikum:
This was merged as 20f6b8894341600b8c025945a13bd6bb6316d9f7.

PR #3603, opened 2020-03-06
Clarify what kind of variable we're talking about

Newbytee:
Make it clearer that you should set an environment variable and not something else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,9 @@ targets to build Proton:
 `make install` - This will install Proton into your user's Steam directory.
 You may need to restart the Steam client to see it. The tool's name in the
 Steam client will be based on the currently checked out branch of Proton. You
-can override this name using the `build_name` variable.
+can override this name using the `build_name` make variable, which can be set 
+via an environment variable or an argument to the make command 
+[among other ways](https://www.gnu.org/software/make/manual/html_node/Values.html#Values).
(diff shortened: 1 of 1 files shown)


aeikum:
Sort of. It's actually a make variable, so you can also specify it like `make build_name=blah redist`. As you say, you can also set it in the environment. There's more info about how make fills in its variables here: https://www.gnu.org/software/make/manual/html_node/Values.html#Values

The `help` target in the Makefile makes this a little more clear, maybe there's a good way to update the `README` to match.

Newbytee:
How does that sound?

README.md
```diff
@@ -66,7 +66,9 @@ targets to build Proton:
 `make install` - This will install Proton into your user's Steam directory.
 You may need to restart the Steam client to see it. The tool's name in the
 Steam client will be based on the currently checked out branch of Proton. You
-can override this name using the `build_name` variable.
+can override this name using the `build_name` make variable, which can be set 
+via an evironment variable or an argument to the make command
```
HonkingGoose: There's a small typo.
Change `eviroment` to `environment`.

Newbytee: Thank you, I'll fix that.

HonkingGoose: Cheers! 😄

PR #3664, opened 2020-03-20
build: Use MAKE variable for building modules.

zzhiyi:
So that options like -j can be passed through.

Signed-off-by: Zhiyi Zhang <zzhang@codeweavers.com>


aeikum:
Looks good, will merge.

aeikum:
This was merged as 60b53f915d76cf969dd16e740d6683c4330e4041.

PR #3826, opened 2020-05-03
gstreamer: Add orc to enable building gstreamer with orc.

GloriousEggroll:
This fixes slow-downs with video playback on lower end machines.

There was some discussion about this over IRC a few days ago. The root cause of the problem is that steam runtime ships orc 0.4.16, which is too old for the current gstreamer implementation:
```
Dependency orc-0.4 found: NO found '0.4.16' but need: '>= 0.4.24'
```
steam runtime has 0.4.16:
http://repo.steampowered.com/steamrt/pool/main/o/orc/

however orc is necessary for video playback, otherwise you hit issues like this in RE2:
https://github.com/Guy1524/wine/issues/1

This PR adds orc 0.4.31 to the build environment and allows gstreamer to utilize it during build time.

Some notes:

1) I had to add LD_LIBRARY_PATH lines to the ninja build lines otherwise it would not recognize the libraries for some reason, even though pkgconfig was set and was being picked up:

```
/home/vagrant/build-Proton-5.7-GE-1-MF/obj-tools32/bin/orcc: symbol lookup error: /home/vagrant/build-Proton-5.7-GE-1-MF/obj-tools32/bin/orcc: undefined symbol: orc_version_string
```
2) I did not use GST_COMMON_MESON_ARGS for orc because it disables tools by default, and we need tools enabled for orcc to be generated in tool32/bin.

### DIFF
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -198,6 +198,10 @@ GLIB := $(SRCDIR)/glib
 GLIB_OBJ32 := ./obj-glib32
 GLIB_OBJ64 := ./obj-glib64
 
+GST_ORC := $(SRCDIR)/gst-orc
+GST_ORC_OBJ32 := ./obj-gst-orc32
+GST_ORC_OBJ64 := ./obj-gst-orc64
+
 GSTREAMER := $(SRCDIR)/gstreamer
 GSTREAMER_OBJ32 := ./obj-gstreamer32
 GSTREAMER_OBJ64 := ./obj-gstreamer64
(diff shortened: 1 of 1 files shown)


build/makefile_base.mak
```diff
@@ -294,6 +298,7 @@ FONTS_OBJ := ./obj-fonts
 OBJ_DIRS := $(TOOLS_DIR32)        $(TOOLS_DIR64)        \
             $(FFMPEG_OBJ32)       $(FFMPEG_OBJ64)       \
             $(GLIB_OBJ32)         $(GLIB_OBJ64)         \
+            $(GST_ORC_OBJ32)    $(GST_ORC_OBJ64)       \
```
jqadev, line 301: Spaces in the second and third columns are not adjusted.

build/makefile_base.mak
```diff
@@ -835,13 +924,17 @@ gst_good: gst_good32 gst_good64
 
 gst_good64: SHELL = $(CONTAINER_SHELL64)
 gst_good64: $(GST_GOOD_CONFIGURE_FILES64)
-	PATH="$(abspath $(TOOLS_DIR64))/bin:$(PATH)" ninja -C "$(GST_GOOD_OBJ64)" install
+	PATH="$(abspath $(TOOLS_DIR64))/bin:$(PATH)" \
+    LD_LIBRARY_PATH="$(abspath $(TOOLS_DIR64))/lib:$(LD_LIBRARY_PATH)" \
```
jqadev, line 928: Missing alignment with other lines.

jqadev: You use both spaces and tabs. Spaces are preferred. You can set usage of tab as 4 spaces in your IDE.

aeikum:
Hey @GloriousEggroll . I cleaned up your commit a bit and it's now in the proton_5.0-next branch.

aeikum:
This was merged in 5.0-8 as b6b52d19f115c8457f773f53d06cab7d6400cc88. Thanks!

PR #3845, opened 2020-05-08
Avoid unnecessary dist.lock creation in system-wide Proton installs

### LINKED ISSUE #3844: Ensure dist.lock management does not break system-wide installs
# Feature Request

## I confirm:
      contain this feature already.

## Description

The `proton` python script currently creates a lock file based on the `base_dir` location. This works ok for "mainstream" Proton installs, as they get installed by the Steam client in the SteamApps directory where the user must have write access to; however, this can break in Proton versions installed under system-wide compatibilitytools.d directories such as `/usr/share/steam` or `/usr/local/share/steam`, because there the user could have no write access to create that lock file.
A solution could be to check for write access in `base_dir` before attempting to create the lock file in there, resorting to another writable location or to directly use the `/var/lock` directory in all cases.
Using a directory other than `base_dir` poses the problem of having a unique ID for the lock file at one's disposal, but I think this can be done by using the `CURRENT_PREFIX_VERSION` variable.

## Justification

This issue is for the `proton` script, that is not part of standard wine

## Risks

Choosing a bogus ID for the lock file could mean possible concurrent access to the prefix that should have been blocked by the lock file.

## References
https://github.com/ValveSoftware/steam-for-linux/issues/6310

414n:
Closes #3844

### DIFF
--- a/proton
+++ b/proton
@@ -109,9 +109,7 @@ class Proton:
 
     def extract_tarball(self):
         with self.dist_lock:
-            if not os.path.exists(self.dist_dir) or \
-                    not os.path.exists(self.path("dist/version")) or \
-                    not filecmp.cmp(self.version_file, self.path("dist/version")):
+            if self.need_tarball_extraction():
                 if os.path.exists(self.dist_dir):
                     shutil.rmtree(self.dist_dir)
                 tar = None
@@ -126,15 +124,25 @@ class Proton:
                 tar.close()
                 try_copy(self.version_file, self.dist_dir)
 
+    def need_tarball_extraction(self):
+        '''Checks if the proton_dist tarball archive must be extracted. Returns true if extraction is needed, false otherwise'''
+        return not os.path.exists(self.dist_dir) or \
+            not os.path.exists(self.path("dist/version")) or \
+            not filecmp.cmp(self.version_file, self.path("dist/version"))
+
     def make_default_prefix(self):
         with self.dist_lock:
             local_env = dict(g_session.env)
-            if not os.path.isdir(self.default_pfx_dir):
+            if self.missing_default_prefix():
                 #make default prefix
                 local_env["WINEPREFIX"] = self.default_pfx_dir
                 local_env["WINEDEBUG"] = "-all"
                 g_session.run_proc([self.wine_bin, "wineboot"], local_env)
                 g_session.run_proc([self.wineserver_bin, "-w"], local_env)
+
+    def missing_default_prefix(self):
+        '''Check if the default prefix dir is missing. Returns true if missing, false if present'''
+        return not os.path.isdir(self.default_pfx_dir)
 
 class CompatData:
     def __init__(self, compatdata):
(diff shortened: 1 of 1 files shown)


aeikum:
Hmm. I don't think this is the right solution. For a system installation like that, the dist/ directory should already be extracted.

What about this: do the version check before the lock, grab the lock only if that fails, re-check the version to ensure someone didn't beat us to it, and proceed with the extraction if that 2nd check fails.

414n:
Thanks for taking the time to look into this, @aeikum 

I agree that, for now, the current solution is just a stopgap that I've put in place to make things work on my system.

If I understood you correctly, I think there's still an issue with the double-check strategy that you're proposing: the lock is still initialized inside the Proton class constructor, so in a system-wide install it will throw an exception as soon as an instance is initialized.
I think we'd need a lazy `FileLock` initialization to implement that or a 100% safe location where the lock won't ever have permission/access issues.

aeikum:
> If I understood you correctly, I think there's still an issue with the double-check strategy that you're proposing: the lock is still initialized inside the Proton class constructor, so in a system-wide install it will throw an exception as soon as an instance is initialized.

Are you sure? Looking at the filelock implementation, I think it only creates the file when acquiring the lock. The class initializers only store the path string.

414n:
You're right, the error trace is indeed this:

```
  File "/usr/share/steam/compatibilitytools.d/Proton-5.8-GE-1-MF/proton", line 679, in <module>
    g_proton.extract_tarball()
  File "/usr/share/steam/compatibilitytools.d/Proton-5.8-GE-1-MF/proton", line 112, in extract_tarball
    with self.dist_lock:
  File "/usr/share/steam/compatibilitytools.d/Proton-5.8-GE-1-MF/filelock.py", line 323, in __enter__
    self.acquire()
  File "/usr/share/steam/compatibilitytools.d/Proton-5.8-GE-1-MF/filelock.py", line 271, in acquire
    self._acquire()
  File "/usr/share/steam/compatibilitytools.d/Proton-5.8-GE-1-MF/filelock.py", line 384, in _acquire
    fd = os.open(self._lock_file, open_mode)
PermissionError: [Errno 13] Permission denied: '/usr/share/steam/compatibilitytools.d/Proton-5.8-GE-1-MF/dist.lock'
```

I hurriedly assumed the issue was in the initialization, but that is not the case.
Given this, I'll rename the PR and send an improved patch with your suggestions.

414n:
Here it is. I refactored the checks into class methods so they can be tested both outside the class and inside the `extract_tarball` and `make_default_prefix` methods.

aeikum:
Thanks, that looks perfect. It also fixes the issue for you, right?

414n:
Yes, I did a test before pushing and it seems to work ok

aeikum:
Great, thanks. Will merge.

414n:
Thanks, I just cleaned up the branch history, keeping only the latest changes.

aeikum:
This was merged in 5.0-8 as 6e821c774fce67e35cedc2d193853c9333a1922f. Thanks!

Issue #3788, opened 2020-04-21
Mod and Play (1127460)


Meister1593:
# Compatibility Report
- Name of the game with compatibility issues: Mod & Play 
**(Unreal engine 4 game)**

- Steam AppID of the game: 1127460

## System Information
- Distribution: Arch rolling
- GPU: Gigabyte GTX 1060 (3 GB)
- Driver/LLVM version: nvidia vulkan 440.66.09
- Kernel version: 5.6.4-arch1-1
- Link to full system information report as [Gist](https://gist.github.com/Meister1593/ca6f8d4c0ebf69dbe00e83b916f7ec08)
- Proton version: Basically any, 4.11-13, 5.0-6, 5.4-GE-3, 5.5-GE-1, 5.6-GE-2

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

[steam-1127460.log](https://github.com/ValveSoftware/Proton/files/4511781/steam-1127460.log)

## Symptoms 

Game does launch okay, but loading basically any level, including in multiplayer causes game to drop back to menu. It also shows level for a fraction of a second, and instantly drops to menu.
Disabling esync, fsync as well as changing driver to usual (not vulkan branch) does not help. 
Enabling `PROTON_USE_WINED3D` does not help too.
**Interesting thing that game did run before last engine update (i think it's important)**

## Reproduction

1. Launch game
2. Try to start or join game

kisak-valve:
Possible line of interest from the log: `err:steamclient:create_win_interface Don't recognize interface name: SteamNetworkingSockets004`

Meister1593:
> Possible line of interest from the log: `err:steamclient:create_win_interface Don't recognize interface name: SteamNetworkingSockets004`

Yeah i noticed that too, but i don't see how this could be resolved.
This could be resolved manually somehow?

Yogfan800:
I have found 2 ways to get single player to work. Only downside is that workshop items don't work.

1) Run game with Proton 4.2-9
2) Run ModandPlay-Win64-Shipping.exe located in the games installation folder at /modandplay/modandplay/binaries/Win64.

It definitely seems like the SteamNetworkingSockets004 thing you guys mention could be causing the issue.

stevenlafl:
> Possible line of interest from the log: `err:steamclient:create_win_interface Don't recognize interface name: SteamNetworkingSockets004`

[...]

PR #3887, opened 2020-05-17
Implement SteamNetworkingSockets004

stevenlafl:
In reference to https://github.com/ValveSoftware/Proton/issues/3788 I have implemented SteamNetworkingSockets004 and it appears to work (see https://github.com/ValveSoftware/Proton/issues/3788#issuecomment-629769103)

I noticed that SDK 146's headers has SteamNetworkingSockets003 inside of it. This is not consistent with Steamworks SDK 146, as those headers show SteamNetworkingSockets004 listed as the version bundled.

SteamNetworkingSockets003 and SteamNetworkingSockets004 appear to have the same basic signature according to the headers.

A compiled version based on Proton 5.0-7 is here https://github.com/stevenlafl/Proton/releases/tag/proton-5.0-7-sockets004


stevenlafl:
We may be able to get away with a change to `gen_wrapper.py` by adding `SteamNetworkingSockets004` to `aliases` but I'm not sure if it's that simple.

aeikum:
I implemented it differently, but SteamNetworkingSockets004 is supported now in 5.0-8. Valve screwed up the numbering quite badly around this time (one of the older SDKs is straight up wrong), but I was able to verify the correct ABI for each version.

PR #3940, opened 2020-06-05
Update Vagrantfile to change disk size

netbospl:


netbospl:
might need 'vagrant plugin install vagrant-disksize' to be executed before, alto I've done this always before and always I've been executing 'resize-vagrant-disk.sh' script after vagrant up

netbospl:
Updated makefile to execute 'resize-vagrant-disk.sh' on make vagrant

aeikum:
Hey @netbospl . I didn't know about this plugin! It looks useful. I am going to add this to my local tree to make sure it works well on all of our build platforms, and if so, will merge it. Thanks!

aeikum:
Our base VM provider has updated the default disk size to 128 GB, so I think this isn't necessary anymore. See https://github.com/lavabit/robox/issues/93#issuecomment-665697891

Thanks for the suggestion!

PR #4284, opened 2020-10-17
Added standard help output

mahaupt:
Added a standard output when calling ./proton without any args
Prevent out of bounds error

### DIFF
--- a/proton
+++ b/proton
@@ -892,8 +892,12 @@ if __name__ == "__main__":
     g_session.init_session()
 
     #prevent out of bounds error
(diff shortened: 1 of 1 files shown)


aeikum:
That should be `< 2`, right?

Anyway, the proton script isn't meant to be run by itself, it's part of the Steam Play project and won't work standalone unless you really know what you're doing. We could instead print out an error message saying that.

mahaupt:
I changed it to `< 2` and edited the error

PR #4285, opened 2020-10-17
media-converter/videoconv.rs: help with FIXME and use of more functional for

orsenkucher:
Hi, I was traveling across the rusty part of Proton
and noticed those FIXME comments in `videoconv.rs`, which I could totally help with
So, here it is😄

### DIFF
--- a/media-converter/src/videoconv.rs
+++ b/media-converter/src/videoconv.rs
@@ -281,15 +281,11 @@ impl VideoConvState {
             None => {
                 let blank = include_bytes!("../blank.ogv");
 
-                let remaining = blank.len() - offs;
+                let to_copy = std::cmp::min(blank.len() - offs, out.len());
 
-                if out.len() <= remaining {
-                    out.copy_from_slice(&blank[offs..(offs + out.len())]);
-                    Ok(out.len())
-                }else{
-                    (0..remaining).for_each(|i| out[i] = blank[offs + i]);
-                    Ok(remaining)
-                }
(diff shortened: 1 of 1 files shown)


media-converter/src/videoconv.rs
```diff
@@ -216,10 +216,7 @@ impl<'a> Read for PadReader<'a> {
         if to_copy == out.len() {
             out.copy_from_slice(&self.chunk[self.chunk_offs..(self.chunk_offs + to_copy)]);
         }else{
-            /* FIXME: there's probably some cool functional way to do this */
-            for i in 0..to_copy {
-                out[i] = self.chunk[self.chunk_offs + i];
-            }
+            (0..to_copy).for_each(|i| out[i] = self.chunk[self.chunk_offs + i]);
```
darksv: I think the whole `if` can be replaced with just `out[..to_copy].copy_from_slice(&self.chunk[self.chunk_offs..(self.chunk_offs + to_copy)]);` 😄

orsenkucher: done👍

aeikum:
@orsenkucher @darksv Thanks! I pushed one more commit to your branch that fixes up the other chunk to avoid the if-statement as well. Does it look good?

orsenkucher:
@aeikum Yup👍 looks good to me

aeikum:
This was merged as 370dc01d07eba692e6d19ceebc7b9e0da07ad0c2. Thanks!

PR #4292, opened 2020-10-18
proton: make log directory configurable via PROTON_LOG_DIR

### LINKED ISSUE #4265: Allow an alternate directory for log files
# Feature Request

## I confirm:
      contain this feature already.

## Description 

Please add an option that I can set, possibly in `user_settings.py`, to put steam-*.log files somewhere other than $HOME. I'd rather avoid cluttering that directory if possible.

Newbytee:
Makes the path which Proton stores its log at configurable via the environment variable PROTON_LOG_DIR. Note that this changes the directory where Proton stores the log, not the entire path, and as such no filename should be specified. Also worth noting is that this path does not make Proton attempt to create the file path necessary to store a file there; if it fails, it fails. I was considering making it so that Proton attempted to create the entire path by itself if it did not exist, but I reasoned that the added complexity wouldn't be worth the small gain, and a user advanced enough to want to change this likely would be able to create the path on their own. That said, if you prefer it to be more robust, I can change this.

Fixes https://github.com/ValveSoftware/Proton/issues/4265

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,7 +264,8 @@ the Wine prefix. Removing the option will revert to the previous behavior.
 
 | Compat config string  | Environment Variable           | Description  |
 | :-------------------- | :----------------------------- | :----------- |
-|                       | <tt>PROTON_LOG</tt>            | Convenience method for dumping a useful debug log to `$HOME/steam-$APPID.log`. For more thorough logging, use `user_settings.py`. |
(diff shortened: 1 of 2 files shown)


aeikum:
Thanks, will merge.

aeikum:
I included this in the 5.0-10 release (f4ae0879ea3a57df5453f3f493b3c5305f9d641c). I'm also planning to include it in an upcoming 5.13 release. Let's leave this open until that happens.

aeikum:
This was merged into the 5.13 branch as 6305de83bcfb37f965cd52ba109a143694b85907. Thanks!

PR #4343, opened 2020-11-02
Some proposals for improving virtual machine creation

giomasce:
These three more or less independent commits all go in the direction of making virtual machine creation faster or more robust.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -66,6 +66,8 @@ Vagrant.configure(2) do |config|
 
     debian10.vm.provision "shell", privileged: "true", inline: <<-SHELL
       set -e
+      # Uncomment this is you have apt-cacher-ng running on your host machine
+      #export http_proxy="http://192.168.121.1:3142/"
       #install docker and steam-runtime dependencies
       dpkg --add-architecture i386
       apt-get update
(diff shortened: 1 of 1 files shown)


aeikum:
The `set -e` thing is a good idea, but the rest is pretty iffy. `libeatmydata` doesn't look like a great idea (the idea of the Vagrant VM is to have a reliable build environment), and the apt-cache thing requires a lot of setup on the host, which isn't even applicable to many distros.

giomasce:
BTW, I don't think that using `libeatmydata` makes the virtual machine creation any less reliable, unless the virtual machine (or the host machine) crashes during the VM creation (in any of which cases you're probably going to drop the half-baked VM and re-create it anyway). It just defers the moment when the caches are written to the disk, but doesn't change what gets written (and it doesn't impact anything after the creation). However, your choice on any of those.

giomasce:
Uhm, I closed this mistakenly by nuking the branch. It is still valid, actually.

ivyl:
`set -e` is nice and helps to notice problems with VM creation during the VM creation. I've pushed the change to experimental. I am not so sure about the rest. Since the recommended way to build is Podman/Docker I'll just drop them.

Thanks!

giomasce:
Yeah, it's not really relevant any more.

PR #4353, opened 2020-11-06
lsteamclient: Allow callbacks to be NULL

giomasce:
Steamworks API seems to accept NULL callbacks, simply avoiding to call
them when this is the case.


aeikum:
Thanks! Will merge.

PR #4362, opened 2020-11-10
Port lsteamclient/gen_wrapper.py to Python 3

giomasce:
Python 2 is already EOL and will be less and less available in the future. The first patch about making the output more deterministic is to check that Python 3 is not itself changing the output (and, I believe, it has value in itself as well).


aeikum:
Thanks, looks good. Will merge.

aeikum:
These were merged as 04af5e3d79ff27db568005291fd114b3598746a4^..e5c634ce0b4252e231119d87bf174c8194ecd311.

PR #4409, opened 2020-11-21
fix for ValveSoftware/Proton/#4269

gitaen:
symlinks copied to pfx directory were using wrong relative paths

         ./pfx/drive_c/windows/syswow64/kernel32.dll -> dist/lib/wine/fakedlls/kernel32.dll

fixed to use absolute path

### DIFF
--- a/proton
+++ b/proton
@@ -294,7 +294,7 @@ class CompatData:
             if os.path.dirname(contents).endswith(('/lib/wine', '/lib/wine/fakedlls', '/lib64/wine', '/lib64/wine/fakedlls')):
                 # wine builtin dll
                 # make the destination an absolute symlink
-                contents = os.path.normpath(os.path.join(os.path.dirname(src), contents))
+                contents = os.path.abspath(os.path.join(os.path.dirname(src), contents))
             if dll_copy:
                 try_copyfile(src, dst)
             else:


aeikum:
Hmm. I'm not necessarily opposed to this, but I think it shouldn't be necessary. The src argument to that function should be a full path, since Steam launches Proton with a full path. How are you launching Proton so it gets in this state?

gitaen:
I am launching it with
`STEAM_COMPAT_DATA_PATH=/tmp/proton_data ./proton run binary.exe`

F1xteR:
Even with the fix applied Proton 5.13-4 still won't launch, same error code.
Did i do something wrong? i applied the fix in the Proton python file and deleted pfx, pycache and dist folders.
I also launched it like above just setting STEAM_COMPAT_DATA_PATH to steamapps/common/Proton 5.13

ghost:
@F1xteR STEAM_COMPAT_DATA_PATH should be set to the directory where your pfx wineprefix will be. I set it to $HOME/proton

opl-:
In order for this fix to work, you will need to create a new `compatdata` directory (wine prefix) for the game that's crashing. This is because the issue exists within the code that initializes it which won't run if one already exists.

Step by step instructions (for Steam users):

1. Apply this patch (for example by manually editing the `SteamLibrary/steamapps/common/Proton 5.13/proton` file to apply [the change](https://github.com/ValveSoftware/Proton/pull/4409/files)).
2. Find out the app ID of the crashing game through [SteamDB](https://steamdb.info) or (in Steam beta versions) by right clicking it in the Steam library, opening the Properties window and looking in the bottom left corner.
3. Go to `SteamLibrary/steamapps/compatdata/`, rename the directory named after the app ID from step 2 and create a new, empty directory in its place.
    - **This directory might contain your saves or other user data.** It includes directories such as App Data and Documents. Make sure to backup any important files before deleting it.
4. Run the game from your Steam library. The empty directory should now contain a fresh wine prefix.
5. If needed, copy any saves or other user data you need from the old, renamed directory into the new `compatdata` directory and verify that it worked. After that the old, renamed directory can be deleted.

Frigyes06 (approved):

nyabinary:
Why did this never get merged?

PR #4560, opened 2021-01-19
gitmodules: Consistently remove trailing slash

smcv:
The forms with or without a trailing slash are interchangeable via https,
but the form with a trailing slash doesn't work if you have git configured
to always access Github repositories via ssh, using something like:

    git config --global url."git+ssh://git@github.com/".insteadOf "https://github.com/"


aeikum:
Thanks, will merge.

aeikum:
This is now in proton_5.13 and experimental_5.13. Closing.

PR #4562, opened 2021-01-20
(issue not a pull request) Skyrim SE, all other Bethesda games crash on launch when opening with proton GE builds

minusfourty:
I tried following the manual setup from the readme.md, and after installing Proton GE 6.0 and all of its dependencies, but the game refuses to open at all. After hitting the play button, the steam icon spins for a bit, and nothing happens. How am I supposed to get this to work? Doom runs with it just fine, but for whatever reason elder scrolls and fallout games won't launch at all. Any help would be much appreciated


kisak-valve:
Hello @minusfourty, we're using one issue report per unofficially supported game title, so I've gone ahead and transferred this issue report to https://github.com/ValveSoftware/Proton/issues/4#issuecomment-763852773, but really, this feedback should be given to the third party build of Proton. Additionally, please do not abuse pull requests as issue reports.

PR #4729, opened 2021-04-02
Cleanup toplevel Makefile rules, configure.sh usage and CCACHE environment.

rbernon:


aeikum:
This is in current experimental_6.3.

aeikum:
Actually I think `build: Pass -j flags and overrides to sub-make inside Vagrant.` breaks something. For me, it causes the `drive_c/openxr` directory not to exist within the default prefix. I think something is happening out of order.

Instead, I squashed these commits into earlier commits and backed out the other changes until we figure this out:

    bf53c0e3 build: Fix install rule folder cleanup.
    7e55c34d build: Fix the individual module install directory.
    6bb944b4 build: Install redist to REDIST_DIR, not DEPLOY_DIR.

rbernon:
Added two commits to fix the broken parallel(er) build:

build: Install wineopenxr64.json after the prefix is created. 
build: Simplify and cleanup TTF font creation rules.

ivyl:
Hey, I've tried to give this a shot, as some distros have already updated to Ruby 3.0 which breaks Vagrant...

I've been getting:
`Error: container_linux.go:367: starting container process caused: chdir to cwd ("/home/ivyl/src/build-6.3-2") set in config.json failed: permission denied: OCI permission denied`

The error is not very obvious but I dug around and it turns out that the problematic part is `-v $(HOME):$(HOME)`. Shadowing a whole home directory is apparently not allowed by [OCI](https://opencontainers.org/) and I think it's a good thing. Poking around it looks unnecessary, as we already have `-v $(SRC):$(SRC) -v $(OBJ):$(OBJ)`.

On top of that, if we want to use the rootless mode of [docker](https://docs.docker.com/engine/security/rootless/) or [podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) the UID shenanigans (`-e USER -e USERID=$(shell id -u) -u $(shell id -u):$(shell id -g)`) won't work.

Rootless mode uses [user_namespaces](https://man7.org/linux/man-pages/man7/user_namespaces.7.html) which map the pseudo root (uid 0) in the container to your `id -u` outside. Everything else is mapped to your user's [subuids and subgids](https://www.man7.org/linux/man-pages/man5/subuid.5.html).

Giving an user uncontrolled access to root containers is risky and basically equals to giving them root privileges outside (`-v /etc`?). Also mounting the whole home directory is the container is just scary.

A few ideas:
 * contain everything under Proton's src directory so only `-v $HOME/src/proton` is necessary
 * `build/` in .gitignore, it would contain ccache, and the `OBJ` dirs per name build
 * build everything in the container, this includes the afdko invocations and making sure we have the shared dist downloads in place (but you can also `make downloads` outside if you don't want to have networked container).
 * make it so that `docker run -v ${PROTON_DIR}:${PROTON_DIR} -w ${PROTON_DIR} ${PROTON_SDK} make dist` is enough to get everything built
 * provide helpers for both rootless and root modes can be easily invoked without the full command line

If PROTON_SDK is not enough to build everything, we may want to introduce an image built on top of it that contains all the extra dependencies.

rbernon:
> Hey, I've tried to give this a shot, as some distros have already updated to Ruby 3.0 which breaks Vagrant...
> 
> I've been getting:
> `Error: container_linux.go:367: starting container process caused: chdir to cwd ("/home/ivyl/src/build-6.3-2") set in config.json failed: permission denied: OCI permission denied`
> 
> The error is not very obvious but I dug around and it turns out that the problematic part is `-v $(HOME):$(HOME)`. Shadowing a whole home directory is apparently not allowed by [OCI](https://opencontainers.org/) and I think it's a good thing. Poking around it looks unnecessary, as we already have `-v $(SRC):$(SRC) -v $(OBJ):$(OBJ)`.
> 
> On top of that, if we want to use the rootless mode of [docker](https://docs.docker.com/engine/security/rootless/) or [podman](https://github.com/containers/podman/blob/master/docs/tutorials/rootless_tutorial.md) the UID shenanigans (`-e USER -e USERID=$(shell id -u) -u $(shell id -u):$(shell id -g)`) won't work.
> 

Yes, this was what was done at some point, but there were permissions issues when folders aren't previously created and when docker creates them itself (as root). This was the case for instance for $HOME/.ccache (that was previously shared too).

> Rootless mode uses [user_namespaces](https://man7.org/linux/man-pages/man7/user_namespaces.7.html) which map the pseudo root (uid 0) in the container to your `id -u` outside. Everything else is mapped to your user's [subuids and subgids](https://www.man7.org/linux/man-pages/man5/subuid.5.html).
> 
> Giving an user uncontrolled access to root containers is risky and basically equals to giving them root privileges outside (`-v /etc`?). Also mounting the whole home directory is the container is just scary.
> 

Sure, this is also something that was previously there but temporarily dropped for the reason above.

> A few ideas:
> 
>     * contain everything under Proton's src directory so only `-v $HOME/src/proton` is necessary
> 
>     * `build/` in .gitignore, it would contain ccache, and the `OBJ` dirs per name build
> 

Sharing ccache as much as possible makes it more powerful, and keeps the cache management simple. IMHO keeping it where the user has set it is better than making it local. The default setting is a cache of 5G for instance, and that would make every proton build tree take up to 5G more on disk.

>     * build everything in the container, this includes the afdko invocations and making sure we have the shared dist downloads in place (but you can also `make downloads` outside if you don't want to have networked container).
> 

This is the intention, but there are currently some unsolved issues. The steam runtime base container doesn't contain the required python packages for instance (it doesn't have pip and most debian python packages have been removed from the distribution).

>     * make it so that `docker run -v ${PROTON_DIR}:${PROTON_DIR} -w ${PROTON_DIR} ${PROTON_SDK} make dist` is enough to get everything built
> 

Not sure to understand, do you mean running that command outside of make? Why ask the user to bother with docker / vagrant setups? I'd rather have to type `make dist` only and have everything built, which is already what we have with Vagrant, and this is trying to achieve a simple way to build without it.

>     * provide helpers for both rootless and root modes can be easily invoked without the full command line
> 

As indicated above, I intended to drop the UID thing at some point. I also have local patches to add an option to chose podman instead, however I had trouble with it where sometimes the rsync changes were not visible inside the container for some reason. I think it has some overlayfs cache or something that sometimes break so it's not very reliable.

> 
> If PROTON_SDK is not enough to build everything, we may want to introduce an image built on top of it that contains all the extra dependencies.

We can add things to the SDK image easily, but we have to keep the license constraints in mind as we redistribute software in binary form. Anything we add has to be also published as source (so using steam runtime packages is better if possible).

ivyl:
> Yes, this was what was done at some point, but there were permissions issues when folders aren't previously created and when docker creates them itself (as root). This was the case for instance for $HOME/.ccache (that was previously shared too).

That's seems overly heavy-handed for something that can be solved by `mkdir -p` before the container is started.

> Sharing ccache as much as possible makes it more powerful, and keeps the cache management simple. IMHO keeping it where the user has set it is better than making it local. The default setting is a cache of 5G for instance, and that would make every proton build tree take up to 5G more on disk.

I've meant sharing it for all proton builds but keeping it inside of the proton dir. Something like:

```
build/ccache
build/6.3-2-myfix1
build/6.3-experimental-myfix2
```
Which would be an improvement over current situation where the build dir is places at the same level as proton source directory... which frankly caught me off guard.

I don't think sharing ccache outside of the proton build context would help with anything and IMO it's not worth mounting the whole `$HOME`.

> Not sure to understand, do you mean running that command outside of make? Why ask the user to bother with docker / vagrant setups? I'd rather have to type make dist only and have everything built, which is already what we have with Vagrant, and this is trying to achieve a simple way to build without it.

I mean it as a transparency measure. Right now it's not that easy to find what exactly gets executed and how the container is set up. The convenience Makefile rules you prefer to use would become a single docker invocation that's easy to track down and understand.

Also I prefer to start the container myself, but I am more distrustful than most :-)

> As indicated above, I intended to drop the UID thing at some point. I also have local patches to add an option to chose podman instead, however I had trouble with it where sometimes the rsync changes were not visible inside the container for some reason. I think it has some overlayfs cache or something that sometimes break so it's not very reliable.

Rsync changes were not visible? Is this for docker-in-vagrant scenario? This may be interesting to debug.

rbernon:
> > Yes, this was what was done at some point, but there were permissions issues when folders aren't previously created and when docker creates them itself (as root). This was the case for instance for $HOME/.ccache (that was previously shared too).
> 
> That's seems overly heavy-handed for something that can be solved by `mkdir -p` before the container is started.
> 

Sure, it was just part of a commit that have been reverted without thinking much more about it.

It's not part of this series either, as this is only trying to make opting-out of Vagrant more transparent (and using docker directly is already possible by manually calling configure.sh / make).

Making rootless docker usage possible and reducing the scope of the mounted directories could be another PR.

> > Sharing ccache as much as possible makes it more powerful, and keeps the cache management simple. IMHO keeping it where the user has set it is better than making it local. The default setting is a cache of 5G for instance, and that would make every proton build tree take up to 5G more on disk.
> 
> I've meant sharing it for all proton builds but keeping it inside of the proton dir. Something like:
> 
> ```
> build/ccache
> build/6.3-2-myfix1
> build/6.3-experimental-myfix2
> ```
> 
> Which would be an improvement over current situation where the build dir is places at the same level as proton source directory... which frankly caught me off guard.
> 
> I don't think sharing ccache outside of the proton build context would help with anything and IMO it's not worth mounting the whole `$HOME`.
> 

I think it has some advantages, especially as the user may have configured it already but doesn't know that Proton is using a different ccache location and config. Sharing the cache across multiple Proton trees, not just several build folders of the same tree, is also interesting. But, yes, it doesn't need to mount all HOME for that and I intend to change this ultimately.

> > Not sure to understand, do you mean running that command outside of make? Why ask the user to bother with docker / vagrant setups? I'd rather have to type make dist only and have everything built, which is already what we have with Vagrant, and this is trying to achieve a simple way to build without it.
> 
> I mean it as a transparency measure. Right now it's not that easy to find what exactly gets executed and how the container is set up. The convenience Makefile rules you prefer to use would become a single docker invocation that's easy to track down and understand.
> 
> Also I prefer to start the container myself, but I am more distrustful than most :-)
> 

I think we have Vagrant for that purpose, it's better isolated than what docker could ever provide.

> > As indicated above, I intended to drop the UID thing at some point. I also have local patches to add an option to chose podman instead, however I had trouble with it where sometimes the rsync changes were not visible inside the container for some reason. I think it has some overlayfs cache or something that sometimes break so it's not very reliable.
> 
> Rsync changes were not visible? Is this for docker-in-vagrant scenario? This may be interesting to debug.

Didn't try with Vagrant at that time, but I suppose it would be the same. It was an issue when using podman. It's not happening often but it gets annoying when it does, as I had to completely remove the build folder to fix it.

ivyl:
> Making rootless docker usage possible and reducing the scope of the mounted directories could be another PR.

Yep! The changes here are definitely in the right direction and I don't want to block them.

> I think it has some advantages, especially as the user may have configured it already but doesn't know that Proton is using a different ccache location and config. Sharing the cache across multiple Proton trees, not just several build folders of the same tree, is also interesting. But, yes, it doesn't need to mount all HOME for that and I intend to change this ultimately.

Those are fairly advanced topics, and you don't get that with the current Vagrant setup either. TBH I wouldn't expect a contenerized/vm-ized build to fiddle with anything in my `$HOME`. I would even go as far as considering this an anti-pattern.

Having this as an opt-in sounds fine though.

> I think we have Vagrant for that purpose, it's better isolated than what docker could ever provide.

Is this really the reason? I thought it just predates contenerized runtime + probably was something that people introducing it already knew.

IMO ideal solution would allow you building in three ways:
 * single docker/podman invocation with only one `--volume` + one make invocation inside
 * make wrapper that automates that for you, prioritizing rootless solutions if possible and maybe sets some opt-in features
 * Vagrant for people who already have that working and don't want to fiddle with setting up container management system, but that would boil down to `vagrant up && vagrant ssh "option number 2"`

The hybrid solution where some things are build inside and some outside the container just make things confusing.

Also, I am not afraid of people sneaking in exploits, I am afraid of an honest mistake that will result in mounting too much and overwriting my configs or nuking my home dir.

> Didn't try with Vagrant at that time, but I suppose it would be the same. It was an issue when using podman. It's not happening often but it gets annoying when it does, as I had to completely remove the build folder to fix it.

Ah. Since you've mentioned rsync I thought it may be related to Vagrant.

rbernon:
> > Making rootless docker usage possible and reducing the scope of the mounted directories could be another PR.
> 
> Yep! The changes here are definitely in the right direction and I don't want to block them.
> 
> > I think it has some advantages, especially as the user may have configured it already but doesn't know that Proton is using a different ccache location and config. Sharing the cache across multiple Proton trees, not just several build folders of the same tree, is also interesting. But, yes, it doesn't need to mount all HOME for that and I intend to change this ultimately.
> 
> Those are fairly advanced topics, and you don't get that with the current Vagrant setup either. TBH I wouldn't expect a contenerized/vm-ized build to fiddle with anything in my `$HOME`. I would even go as far as considering this an anti-pattern.
> 
> Having this as an opt-in sounds fine though.
> 

I don't think the containerized/vm-ized aspect of the build is anything that matters here, to me it's an implementation detail. Building a C/C++ project that supports ccache, I would expect it to respect my CCACHE_DIR environment and the configuration I've put there.

> > I think we have Vagrant for that purpose, it's better isolated than what docker could ever provide.
> 
> Is this really the reason? I thought it just predates contenerized runtime + probably was something that people introducing it already knew.
> 
> IMO ideal solution would allow you building in three ways:
> 
>     * single docker/podman invocation with only one `--volume` + one make invocation inside
> 
>     * make wrapper that automates that for you, prioritizing rootless solutions if possible and maybe sets some opt-in features
> 
>     * Vagrant for people who already have that working and don't want to fiddle with setting up container management system, but that would boil down to `vagrant up && vagrant ssh "option number 2"`
> 

It's actually kind of already possible, although I didn't try all the combinations and I'm not sure they all work fine.

For 1), for instance, the internal container isn't used if STEAMRT_IMAGE make variable is empty, which is the case if the build directory was configured using `configure.sh --no-steam-runtime`. In that case, you're on your own to provide anything that the build process expects to find in the container (as well as to build the targets usually built outside of it).

For 2), it's also already possible, by using `configure.sh --steam-runtime-image=... --steam-runtime=soldier`, this PR only makes it easier to using it directly from the toplevel Makefile.

For 3), having the user call vagrant to build is really inconvenient, and that was the whole purpose of the top level Makefile, as far as I understand.

Then, as we're building Proton all the time, I personally find that being able to just call `make build_name=... ` from the toplevel, and have a new build tree configured and built is much more convenient than having to manage separate build directories, configure them with `configure.sh ...`, and then call `make`, so I thought that extending the Vagrant build convenience of 3) to 2) was nice to have, hence this PR. Especially as the configure.sh invocation is going to be pretty much always the same.

However, and as @aeikum told me previously, this was not the intended usage of the top-level Makefile, so if it causes more confusion I think I can just keep it as a local change. We would just need to move a few defaults to configure.sh instead of having them in the toplevel Makefile. (And I still think it would make opting-out of Vagrant pretty much non-obvious, as seeing a toplevel Makefile makes people think they are supposed to use make directly to build, and not anything else).

> 
> The hybrid solution where some things are build inside and some outside the container just make things confusing.
> 
> Also, I am not afraid of people sneaking in exploits, I am afraid of an honest mistake that will result in mounting too much and overwriting my configs or nuking my home dir.
> 

Sure, I understand the concerns and ultimately I think we should only mount $(OBJ), $(SRC) and $(CCACHE_DIR).

rbernon:
I actually dropped the vagrant_disable thing, as it looks like it's causing some confusion. I however cleaned up the defaults so that `configure.sh` uses the correct Proton SDK image by default, and the toplevel Makefile gets it from it too instead of hardcoding it.

rbernon:
What this means is that the three ways of building Proton described above should be kind of supported:

1) Building using Vagrant, by running `make` from the source directory, with the targets, options, and defaults it provides on top of the `configure.sh` defaults.

2) Building without Vagrant but with the Proton SDK image, with a few things being built directly on the host (fonts, downloads, and dist targets), by running `configure.sh` from a build directory, then `make`. It has a different set of defaults, and for instance `ccache` use is disabled by default unless `ENABLE_CCACHE=1` is passed to `make`. The Proton SDK image used is either the one set by default in `configure.sh`, or it's also possible to override it with `--proton-sdk-image=...` option.

3) Building without Vagrant or the Proton SDK image (no container involved at all), is the least supported and least tested variant, where you are on your own to provide anything that are expected to build the modules (either on the host or by running `make` within your own containers), and is possible with the same setup as 2), by calling `configure.sh --no-proton-sdk`, then `make`.

ivyl:
I like this much more than using the top level make file. Thanks @rbernon!

It's still a bit on the quirky side but there's no escape from this without major, breaking changes to Proton's build system. It could use some documentation though, especially around different modes of building and all the flags.

I've also added this snippet to be able to build with rootless docker or podman (used via podman-docker).

```diff
diff --git a/build/makefile_base.mak b/build/makefile_base.mak
index 36e8753..769156b 100644
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -55,8 +55,12 @@ else
 	DOCKER_OPTS := $(CCACHE_ENV) -e CCACHE_DISABLE=1 $(DOCKER_OPTS)
 endif
 
-DOCKER_BASE = docker run --rm -e HOME -e USER -e USERID=$(shell id -u) -u $(shell id -u):$(shell id -g) \
-                -v $(SRC):$(SRC) -v $(OBJ):$(OBJ) -w $(OBJ) -e MAKEFLAGS \
+ifneq ($(ROOTLESS_CONTAINER),1)
+	DOCKER_OPTS := -e HOME -e USER -e USERID=$(shell id -u) -u $(shell id -u):$(shell id -g) $(DOCKER_OPTS)
+endif
+
+DOCKER_BASE = docker run --rm -v $(SRC):$(SRC) -v $(OBJ):$(OBJ) \
+                -w $(OBJ) -e MAKEFLAGS \
                 $(DOCKER_OPTS) $(STEAMRT_IMAGE)
 
 STEAMRT_NAME ?= soldier
```

aeikum:
I think this is all in 6.3 now, right?

ivyl:
Yes.

PR #4732, opened 2021-04-03
Fixes some clippy lints in media-converter/

legendofmiracles:
I heard of the rust part in Proton, and had to take a look... 
Just by chance I ran clippy and almost spilled my coffee when i saw the 30 issues! 
This isn't a real performance improvement, it only makes the code more readable.

### DIFF
--- a/media-converter/src/audioconv.rs
+++ b/media-converter/src/audioconv.rs
@@ -217,7 +217,7 @@ struct NeedTranscodeHead {
 
 impl NeedTranscodeHead {
     fn new_from_caps(caps: &gst::CapsRef) -> Result<Self, gst::LoggableError> {
-        let s = caps.get_structure(0).ok_or(gst_loggable_error!(CAT, "Caps have no WMA data!"))?;
+        let s = caps.get_structure(0).ok_or_else(|| gst_loggable_error!(CAT, "Caps have no WMA data!"))?;
 
         let wmaversion = s.get_some::<i32>("wmaversion").map_err(|_| gst_loggable_error!(CAT, "Caps have no wmaversion field"))?;
         let bitrate = s.get_some::<i32>("bitrate").map_err(|_| gst_loggable_error!(CAT, "Caps have no bitrate field"))?;
@@ -227,7 +227,7 @@ impl NeedTranscodeHead {
         let depth = s.get_some::<i32>("depth").map_err(|_| gst_loggable_error!(CAT, "Caps have no depth field"))?;
         let codec_data_buf = s.get::<gst::Buffer>("codec_data")
             .map_err(|_| gst_loggable_error!(CAT, "Caps have no codec_data field"))?
-            .ok_or(gst_loggable_error!(CAT, "Caps have NULL codec_data field"))?;
+            .ok_or_else(|| gst_loggable_error!(CAT, "Caps have NULL codec_data field"))?;
 
         let mapped = codec_data_buf.into_mapped_buffer_readable().unwrap();
         let mut codec_data = Vec::new();
@@ -311,7 +311,7 @@ impl StreamState {
 
             Ok(LoopState::Looping)
         }else{
-            if self.loop_buffers.len() > 0 {
+            if !self.loop_buffers.is_empty() {
                 /* partial loop, track them and then continue */
                 self.buffers.append(&mut self.loop_buffers);
             }
@@ -324,7 +324,7 @@ impl StreamState {
     }
 
     fn write_to_foz(&self) -> Result<(), gst::LoggableError> {
-        if self.needs_dump && self.buffers.len() > 0 {
+        if self.needs_dump && !self.buffers.is_empty() {
             let mut db = (*DUMP_FOZDB).lock().unwrap();
             let db = match &mut *db {
                 Some(d) => d,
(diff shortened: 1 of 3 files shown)


aeikum:
I don't think the whitespace changes are improvements. There a couple more interesting changes, but they're hard to pull out from the whitespace changes. If you can update without the whitespace changes, we can take a look at those.

legendofmiracles:
Foolish of me to not have ran `cargo fmt` in another commit.

But I'll work this out
ETA: tomorrow

legendofmiracles:
@aeikum is this better?

aeikum:
Yes, that looks great, thanks.

media-converter/src/audioconv.rs
```diff
@@ -217,7 +217,7 @@ struct NeedTranscodeHead {
 
 impl NeedTranscodeHead {
     fn new_from_caps(caps: &gst::CapsRef) -> Result<Self, gst::LoggableError> {
-        let s = caps.get_structure(0).ok_or(gst_loggable_error!(CAT, "Caps have no WMA data!"))?;
+        let s = caps.get_structure(0).ok_or_else(|| gst_loggable_error!(CAT, "Caps have no WMA data!"))?;
```
legendofmiracles, line 220: without ok_or_else it would always run the function, no matter if that case happens or not. Thus taking memory and immediately dropping it again.

media-converter/src/audioconv.rs
```diff
@@ -311,7 +311,7 @@ impl StreamState {
 
             Ok(LoopState::Looping)
         }else{
-            if self.loop_buffers.len() > 0 {
+            if self.loop_buffers.is_empty() {
```
legendofmiracles: This doesn't have iterate

media-converter/src/audioconv.rs
```diff
@@ -510,15 +510,15 @@ impl AudioConvState {
             if let Ok(transcoded_size) = read_fozdb.entry_size(AUDIOCONV_FOZ_TAG_PTNADATA, hash) {
                 /* success */
                 let mut buf = vec![0u8; transcoded_size].into_boxed_slice();
-                if let Ok(_) = read_fozdb.read_entry(AUDIOCONV_FOZ_TAG_PTNADATA, hash, 0, &mut buf, fossilize::CRCCheck::WithoutCRC) {
+                if read_fozdb.read_entry(AUDIOCONV_FOZ_TAG_PTNADATA, hash, 0, &mut buf, fossilize::CRCCheck::WithoutCRC).is_ok() {
```
legendofmiracles, line 513: I'm not sure if _ actually makes rust write to memory, but it's more clear like this anyways

media-converter/src/videoconv.rs
```diff
@@ -431,12 +431,12 @@ impl ElementImpl for VideoConv {
 }
 
 struct StreamSerializer<'a> {
-    stream: &'a Vec<u128>,
+    stream: &'a [u128],
```
legendofmiracles, line 434: Borrowing on a Vec is just a slice, no reason to immediately not use a array then

media-converter/src/videoconv.rs
```diff
@@ -541,8 +541,11 @@ impl VideoConv {
 
         if self.sinkpad.peer_query(&mut query) {
             /* XXX: what? */
-            let res = query.get_result();
-            drop(res);
+            let _ = query.get_result();
+            /* Legend: This copies `res` and only then gets dropped ~ doing nothing 
+             * I'll just leave it commented, as it seems important to understanding 
+             * how the program works*/
+            //drop(res);
```
legendofmiracles: I don't feel like I explained this well enough. 
It copies res, and then it drops that copy, leaving the original intact ~ only a unneeded overhead

aeikum: This one is weird. I don't remember the details, but there was some bizarre breakage if I didn't add the explicit drop() here. Hence my "what?" comment. I kind of suspect it was a compiler bug.

legendofmiracles: Alright, admittedly I didn't test my changes. I trust the rust compiler a bit too much on that probably. I will either test them (what game do I even need to run to test this 🤔 ), or we just leave it there for safety.

media-converter/src/audioconv.rs
aeikum: This one is missing a negation, right? We want to execute the body if it is _not_ empty.

media-converter/src/videoconv.rs
aeikum: Taking another look at this one, we call `get_result` twice. In theory we can just drop all of the LOC before the match, right? E.g.:

```diff
@@ -540,9 +540,6 @@ impl VideoConv {
         let mut query = gst::query::Duration::new(gst::Format::Bytes);
 
         if self.sinkpad.peer_query(&mut query) {
-            /* XXX: what? */
-            let res = query.get_result();
-            drop(res);
             state.upstream_duration = match query.get_result() {
                 gst::format::GenericFormattedValue::Bytes(b) =>
                     *b,
```

And it should be functionally equivalent... right? I think we can try this and see if our QA hits a problem during testing.

legendofmiracles:
```rust
        unsafe {
            let mut fmt = mem::MaybeUninit::uninit();
            let mut pos = mem::MaybeUninit::uninit();

            gst_sys::gst_query_parse_duration(self.0.as_ptr(), fmt.as_mut_ptr(), pos.as_mut_ptr());

            GenericFormattedValue::new(from_glib(fmt.assume_init()), pos.assume_init())
        }
```
This is the code inside the get_results() function. If I'm honest, the function names don't look like they'd have side effects - so I reckon it should be alright removing those.

aeikum:
Hey @legendofmiracles . I forced-pushed a commit to your branch (I hope this was the right thing to do, I'm not used to working on GitHub). It fixes the missing negation mentioned above, and includes the removal of the extra `get_result` call.

I'll merge this commit as-is unless I hear otherwise.

media-converter/src/audioconv.rs
legendofmiracles: I'm blind - i would have felt terrible if this had come upstream

legendofmiracles:
This looks good to me! 

Thank you for taking the time to fix my stupid mistakes, I didn't read your comment 😦 - otherwise I would also have fixed it

aeikum:
No problem, that's what code review is for.  Thanks for your contribution.

legendofmiracles:
I'm glad I could contribute!

ivyl:
Looks like this was forgotten. @aeikum anything against including this now?

aeikum:
Not forgotten, I was waiting for our QA to have another reason to test media-converter stuff. Leave it with me.

ivyl:
This is now included in experimental: https://github.com/ValveSoftware/Proton/commit/fc819e4ab6d7b0975f437c1f94f18e01307b98c0

Thanks for your contribution!

PR #4829, opened 2021-05-16
Version fix.

thaewrapt:
`proton` keeps reporting "updating prefix from xxx to 6.3-2" on a prefix update even for 6.3-4 and current Experimental.

Suggestion: either autogenerate this value or read it from some kind of a repository-wide config.

### DIFF
--- a/proton
+++ b/proton
@@ -20,7 +20,7 @@ from filelock import FileLock
 #To enable debug logging, copy "user_settings.sample.py" to "user_settings.py"
 #and edit it if needed.
 
-CURRENT_PREFIX_VERSION="6.3-2"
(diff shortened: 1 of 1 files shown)


kisak-valve:
Hello @thaewrapt, my understanding is that this is not a bug, the Proton version and the wine prefix version are separate and only bumped if there are changes in the Proton version that require the prefix to go through an update iteration. Some patch set updates do not require changes to the wine prefix on disk.

thaewrapt:
Makes sense, but still confusing for the enduser.

aeikum:
Yeah this has come up a few times. Maybe in the next major version I'll use `6.3-a`, `6.3-b`, etc.

PR #4833, opened 2021-05-18
Fix fallback path

cybik:
The current fallback path concatenates .steam directly after $HOME, which leads to Steam Client files failing to be copied into a fresh prefix.

### DIFF
--- a/proton
+++ b/proton
@@ -433,12 +433,8 @@ class CompatData:
                 self.copy_pfx()
 
             # collect configuration info
-            if "STEAM_COMPAT_CLIENT_INSTALL_PATH" in os.environ:
-                #modern steam client sets this
-                steamdir = os.environ["STEAM_COMPAT_CLIENT_INSTALL_PATH"]
-            else:
-                #linux-only fallback, really shouldn't get here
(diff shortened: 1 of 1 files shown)


cybik:
@kisak-valve this is something I stumbled upon by chance.

aeikum:
Good catch, although I think the right fix is to just remove that if statement and assume `STEAM_COMPAT_CLIENT_INSTALL_PATH` is always available.

cybik:
@aeikum technically I can't really say I disagree, since for now proton kind of safely assumes Steam to be present. Though my opinion is having a fallback is always good.

For clarity, I found this by launching a game payload though Proton "outside" of Steam. It's odd, but it's one way of calling Proton that shouldn't be discarded IMO. (Specifically I'm playing around with out-of-steam script-based launching with MTGA)

aeikum:
Running Proton outside of Steam isn't supported. I'd rather make our code here cleaner by removing the conditional.

cybik:
Updated following guidance!

aeikum:
Looks good, thanks. Will merge.

aeikum:
This was merged as 139ad89ed36a47cab790786b8725e782e1e39911. Thanks!

PR #4861, opened 2021-05-26
Create a wine prefix per user

### LINKED ISSUE #4820: Proton not working when multiple Linux users share a single installation path
I've followed [these instructions on Reddit](https://old.reddit.com/r/linux_gaming/comments/5d25oe/how_do_you_share_a_team_library_between_multiple/da1tdv7/) to allow multiple Linux users to share the same installation folder without dealing with permission issues. That seems to work so far for Linux-native games.

The issue with Proton games is that Wine does not allow using a Wine-Prefix that doesn't belong to the current user.

Example:

Both Steam clients on both Linux accounts are configured to download the games in `/opt/games/Steam`.

User `maykin` installs a Windows game and a Wineprefix in `/opt/games/Steam/steamapps/compatdata/430190/pfx` is created.

User `steamuser` wants to play the same game.
It does not need to install the game again, since `maykin` already installed it.
When attempting to start the game nothing happens (no game & no error message).
When inspecting the logs the following error occur:

```
Installing breakpad exception handler for appid(steam)/version(1618256785)
Proton: Upgrading prefix from 5.13-1 to 6.3-2 (/opt/games/Steam/steamapps/compatdata/430190/)
Proton: Removing stale builtin /opt/games/Steam/steamapps/compatdata/430190/pfx//drive_c/windows/system32/amd_ags_x64.dll
Proton: Removing stale builtin /opt/games/Steam/steamapps/compatdata/430190/pfx//drive_c/windows/syswow64/amd_ags_x64.dll
wine: '/opt/games/Steam/steamapps/compatdata/430190/pfx' is not owned by you
wine: '/opt/games/Steam/steamapps/compatdata/430190/pfx' is not owned by you
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_32/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
wineserver: /opt/games/Steam/steamapps/compatdata/430190/pfx is not owned by you
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
ERROR: ld.so: object '/home/steamuser/.local/share/Steam/ubuntu12_64/gameoverlayrenderer.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
wine: '/opt/games/Steam/steamapps/compatdata/430190/pfx' is not owned by you
```

The relevant part is (so I assume) `wine(server): /opt/games/Steam/steamapps/compatdata/430190/pfx is not owned by you`.

The directory is owned by `maykin` but `steamuser` has read and write access to it but Wine - for security reasons apparently - doesn't care and demands that the prefix is owned by the current user, i.e. `steamuser`.

My suggestion:

1. **Create a new prefix per user**:
Instead of creating the prefix here: `/opt/games/Steam/steamapps/compatdata/<gameid>/pfx` we could place it there: `/opt/games/Steam/steamapps/compatdata/<gameid>/pfx/<userid>`
2. **Use symbolic links so that all prefixes share the same data**:
Put the `device_c` in a separate directory outside the prefix, e.g. in `/opt/games/Steam/steamapps/compatdata/<gameid>/data`, and place a symbolic link in `/opt/games/Steam/steamapps/compatdata/<gameid>/pfx/<userid/device_c` that points to the `data` directory

Maykin-99:
Fixes #4820 by setting the base directory of `CompatData` from `steamapps/compatdata/<steamid>/` to `steamapps/compatdata/<steamid>/<userid>/`  (I'm honestly surprised that it worked without a hitch).

The only issue is that the old files still exist.
Ideally we would move them to the new path.
I assume that would belong in `upgrade_pfx`? Or is it even necessary?

### DIFF
--- a/proton
+++ b/proton
@@ -204,7 +204,8 @@ class Proton:
 
 class CompatData:
     def __init__(self, compatdata):
-        self.base_dir = compatdata + "/"
+        self.base_dir = compatdata + "/" + str(os.getuid()) + "/"
+        os.makedirs(self.base_dir, exist_ok=True)
         self.prefix_dir = self.path("pfx/")
         self.version_file = self.path("version")
         self.config_info_file = self.path("config_info")


VannTen:
How could we make this progress forward ?
I just got it by the exact same problem when trying to make family sharing work.

Nomz420:
Could someone look into this? This fix would prove really useful for us.

CorruptComputer:
Does this break Steam Cloud save backups? Might need some coordination with the Steam folks to make sure this doesn't break existing save backups.

Lissy93 (approved):

jessecambon:
+1 also related to https://github.com/ValveSoftware/steam-for-linux/issues/3942

etyarews:
+1 This would fix so many issues

eilimrib:
Please. This is giving me nightmare having to recompile Proton just to share my games with my wife.

Damglador:
@eilimrib you don't have to compile it. Since it's a change in a simple python file, you can just do the same change in already installed Proton version

Damglador:
Wouldn't it be better to use Steam user ID instead of uid on the system? That would be better if you were to move a library to another system, or just use another Steam account on the same local user.

Damglador:
Perhaps it's better to store prefixes in the local compatdata? Because in case user1 with 1000 and user2 with 1001 on the first system won't get swapped prefixes on a system where user1 is 1001 and user2 is 1000

ghost:
So, I found that **this** is the reason why I cannot let my two sons make use of the same gamedata as my account. Seems like a simple fix.

Damglador:
I would do this as a personal patch:
```
self.base_dir = os.path.join(os.environ["STEAM_COMPAT_CLIENT_INSTALL_PATH"], "steamapps", "compatdata", os.environ["SteamAppId"]) + "/"
```
Or, from my understanding, Steam just could supply a local compatdata directory in `STEAM_COMPAT_DATA_PATH` variable, instead of the one in portable library.

Saiv46 (approved):

etyarews:
I've been thinking about it, and Steam really needs to redo the library side of things. I understand this is probably not the best place for it, but...

Yeah, moving /compatdata/ inside a /userid/ is a necessary step (although I'd rather make it `steamapps/compatdata/<userid>/<steamid>/` instead of `/<steamid>/<userid>/`), but the rest of the library also makes no goddamnit sense.

IMO: 

1. `/compatdata/` shouldn't be on the library folder. The library should only include game data. By putting `/compatdata/` inside the library, you make it really hard to use an external device as a "mobile" library between computers with different users in the same household
2. Steam should use a special group for file permissions on the library itself. It is a pain to set up multiple users sharing the same library because the first user who created the library has file permissions on the library, so other users cannot use it or update it.
3. moving `/compatdata/`  (and possibly `/shadercache/`) outside the library folder means that the entire library could have a "shared" group/user permission, while `/compatdata/`  could have the "Linux user" creator permission, which simplifies things because Wine does not like the virtual `c:/` to have permissions other than the user who is executing it

Damglador:
@etyarews there's https://github.com/ValveSoftware/steam-for-linux/issues/12030 for the first point. It also eliminates the need of doing <userid>/<appid>.

From what I understand from the code, Steam passes the compatdata  location to Proton, so it's an issue with Steam and has to be resolved there.

PR #4878, opened 2021-06-01
proton: Copy DLLs provided by the NVIDIA driver into prefix

liam-middlebrook:
The upcoming NVIDIA 470 driver series will introduce a DLL (nvngx.dll)
for the support of NVIDIA DLSS in Proton. This change adds logic for
discovering the location of DLL files provided by the NVIDIA driver, and
copies them to `C:\Windows\System32\`

Related to: https://github.com/ValveSoftware/wine/pull/111


proton
```diff
@@ -127,6 +138,85 @@ def try_get_game_library_dir():
...
+        return None
+
+    # Grab the contents our of our pointer
+    glx_nvidia_info = cast(glx_nvidia_info_ptr, POINTER(link_map)).contents
+
+    # Decode the path to our library to a str()
+    if glx_nvidia_info.l_name is None:
+        return None
+    try:
+        libglx_nvidia_path = glx_nvidia_info.l_name.decode("utf-8")
```
smcv: Should this be

```suggestion
        libglx_nvidia_path = os.fsdecode(glx_nvidia_info.l_name)
```

to cope with non-UTF-8 locales?

(This is unlikely to matter in practice, because `libGLX_nvidia.so.0` is almost certainly in an all-ASCII path like `/run/host/usr/lib/x86_64-linux-gnu` or `/run/host/usr/lib64`.)

smcv (commented):
Not a maintainer, but I suggested a mechanism similar to this one (in C) on https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/issues/71.

aeikum:
Looks good. Pushed a couple small fixups, including @smcv's suggestion.

aeikum:
Oh, one other thought -- this is intended to be 64-bit-only, right? There is no 32-bit support even on Windows etc?

liam-middlebrook:
> Oh, one other thought -- this is intended to be 64-bit-only, right? There is no 32-bit support even on Windows etc?

Correct. DLSS / NGX only work with 64-bit applications.

aeikum:
It's not a blocker, but pylint has a new warning:

    proton:190:19: W0212: Access to a protected member _handle of a client class (protected-access)

smcv:
> It's not a blocker, but pylint has a new warning:
> 
> ```
> proton:190:19: W0212: Access to a protected member _handle of a client class (protected-access)
> ```

I don't think this is going to be avoidable, if we want to use `dlinfo` from Python and pass a `CDLL` to it. `_handle` [is actually documented](https://docs.python.org/3/library/ctypes.html#ctypes.PyDLL._handle), albeit only for the PyDLL subclass; and in any case Proton runs on a runtime that is under our control (currently Steam Runtime 2 'soldier'), so the version of Python isn't going to get newer without a very good reason. That means it's a bit safer than it would ordinarily be to rely on implementation details.

I suppose you could avoid this with something like:

```
        dlopen_func = libdl.dlopen
        dlopen_func.argtypes = c_char_p, c_int
        dlopen_func.restype = c_void_p
        handle = dlopen_func(b"libGLX_nvidia.so.0", os.RTLD_LAZY)
```

aeikum:
This is included in experimental-6.3-20210602 as 2dda8e4e472b09066069c2f3c2ad1e20ecffc997.

liam-middlebrook:
> This is included in experimental-6.3-20210602 as [2dda8e4](https://github.com/ValveSoftware/Proton/commit/2dda8e4e472b09066069c2f3c2ad1e20ecffc997).

I saw some end-user comments that the latest Proton Experimental is now broken on non-NVIDIA systems after this patch due to an exception being thrown in the new `find_nvidia_wine_dll_dir()` function.

```
  File "/home/edmondo/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 1111, in <module>
    g_session.init_session(sys.argv[1] != "runinprefix")
  File "/home/edmondo/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 940, in init_session
    g_compatdata.setup_prefix()
  File "/home/edmondo/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 697, in setup_prefix
    nvidia_wine_dll_dir = find_nvidia_wine_dll_dir()
  File "/home/edmondo/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 152, in find_nvidia_wine_dll_dir
    libglx_nvidia = CDLL("libGLX_nvidia.so.0")
  File "/usr/lib/python3.7/ctypes/__init__.py", line 356, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libGLX_nvidia.so.0: cannot open shared object file: No such file or directory
```
I'll prepare a fix to handle this, but I can't easily test it.

CME42:
Here on debian sid it dies with the following traceback:
```
  File "/home/cme/.steam/steam/steamapps/common/Proton - Experimental/proton", line 1111, in <module>
    g_session.init_session(sys.argv[1] != "runinprefix")
  File "/home/cme/.steam/steam/steamapps/common/Proton - Experimental/proton", line 940, in init_session
    g_compatdata.setup_prefix()
  File "/home/cme/.steam/steam/steamapps/common/Proton - Experimental/proton", line 697, in setup_prefix
    nvidia_wine_dll_dir = find_nvidia_wine_dll_dir()
  File "/home/cme/.steam/steam/steamapps/common/Proton - Experimental/proton", line 151, in find_nvidia_wine_dll_dir
    libdl = CDLL(find_library("libdl"))
  File "/usr/lib/python3.9/ctypes/util.py", line 341, in find_library
    _get_soname(_findLib_gcc(name)) or _get_soname(_findLib_ld(name))
  File "/usr/lib/python3.9/ctypes/util.py", line 147, in _findLib_gcc
    if not _is_elf(file):
  File "/usr/lib/python3.9/ctypes/util.py", line 99, in _is_elf
    with open(filename, 'br') as thefile:
FileNotFoundError: [Errno 2] No such file or directory: b'liblibdl.a'
```

PR #4880, opened 2021-06-02
pefixup: Add timestamps to PE fixup step

misyltoad:
Needed for some anti cheats to be happy.

Signed-off-by: Joshua Ashton <joshua@froggi.es>


Newbytee:
Out of curiosity: Is there any significance to this specific timestamp, i.e. 1622588288, or does it just need to be any timestamp?

misyltoad:
It needs to be non-zero, that timestamp is the time I made the patch.

I didn't make it get the current time for reproducible builds.

ivyl:
Pushed to experimental, thanks.

Issue #4881, opened 2021-06-02
Latest Proton Experimental Fails To Open


cagelight:
Proton Experimental (Build ID 6802103)

I cannot launch any games with the latest Proton Experimental as I get the following error:
```
Traceback (most recent call last):
  File "/home/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 1111, in <module>
    g_session.init_session(sys.argv[1] != "runinprefix")
  File "/home/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 940, in init_session
    g_compatdata.setup_prefix()
  File "/home/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 697, in setup_prefix
    nvidia_wine_dll_dir = find_nvidia_wine_dll_dir()
  File "/home/.local/share/Steam/steamapps/common/Proton - Experimental/proton", line 152, in find_nvidia_wine_dll_dir
    libglx_nvidia = CDLL("libGLX_nvidia.so.0")
  File "/usr/lib/python3.7/ctypes/__init__.py", line 356, in __init__
    self._handle = _dlopen(self._name, mode)
OSError: libGLX_nvidia.so.0: cannot open shared object file: No such file or directory
```
I have an AMD card and it should not be looking for "nvidia" anything.

Vladimir-A:
I have the same situation.

ehea617:
Also seeing the same issue here on an RX 570 with Mesa 20.3.4. Setting PROTON_ENABLE_NVAPI=0 %command% as a launch option does not help either.

DadSchoorse:
It's a known issue, and should hopefully be resolved soon.
https://github.com/ValveSoftware/Proton/pull/4878#issuecomment-853373269

aeikum:
Thanks for reporting. This should be fixed in an Experimental update that just went live (version `experimental-6.3-20210602b`).

cagelight:
Latest update fixed it for me!

davidgrenier:
Likewise.

ehea617:
The issue is resolved for me after updating. Thanks!

kisak-valve:
Closing as fixed by https://github.com/ValveSoftware/Proton/commit/cc7ad4bd13bb88ce194270908cc5243b8dc0613a.

PR #4882, opened 2021-06-02
proton: Improve find_nvidia_wine_dll_dir() handling of missing files

liam-middlebrook:
Additionally split out the find_library() logic for discovery of libdl
so that the case of a missing libdl (no matter how unlikely) is caught
before attempting to load an erroneous CDLL.

Fixes: #4881 (Latest Proton Experimental Fails To Open)

### DIFF
--- a/proton
+++ b/proton
@@ -148,10 +148,20 @@ def try_get_game_library_dir():
 #
 # On failure, returns None
 def find_nvidia_wine_dll_dir():
-    libdl = CDLL(find_library("libdl"))
-    libglx_nvidia = CDLL("libGLX_nvidia.so.0")
+    try:
+        libdl_soname = find_library("dl")
 
-    if libdl is None or libglx_nvidia is None:
+        # If we couldn't determine the proper soname for libdl, bail out.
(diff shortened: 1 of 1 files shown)


proton
```diff
@@ -148,10 +148,15 @@ def try_get_game_library_dir():
 #
 # On failure, returns None
 def find_nvidia_wine_dll_dir():
-    libdl = CDLL(find_library("libdl"))
-    libglx_nvidia = CDLL("libGLX_nvidia.so.0")
+    try:
+        libdl_soname = find_library("dl")
```
DadSchoorse, line 152: I think you should check for `None` here as well, at least that's what the documentations says will be returned if no library can be found.

liam-middlebrook: Ah yeah, I see that here on a second pass-through: https://docs.python.org/3/library/ctypes.html#finding-shared-libraries

edmondo:
Thanks.
Patch applied to the proton script and it fixes the issue.

aeikum:
Thanks for the quick fix. Will get that out shortly.

PR #4892, opened 2021-06-09
Simplify NVIDIA dll discovery

liam-middlebrook:
After testing DLSS+Proton on a different system than I normally use I discovered that the call to `find_library("dl")` wasn't properly discovering libdl.so.2

Additionally tweak the copy behavior so that if any of the driver-provided DLLs are missing Proton is still able to launch.

### DIFF
--- a/proton
+++ b/proton
@@ -23,7 +23,6 @@ from ctypes import cast
 from ctypes import c_int
 from ctypes import c_char_p
 from ctypes import c_void_p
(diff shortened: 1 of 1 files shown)


aeikum:
This was merged in experimental-6.3-20210616 as f7b8a814cb837b638cea970b2c416524c3cff64a and ae13658e1f8279386a3d2e845a5667ea2566848f.

PR #4634, opened 2021-02-18
Fix build failure on Fedora 33

Endle:
In Fedora 33 and some distros, `/bin` is not part of `$PATH` by default. When launching steam-proton-dev with `-e PATH=$(PATH)`, it would complain `/bin/bash: mkdir: command not found`

Signed-off-by: Zhenbo Li <litimetal@gmail.com>

PR #4935, opened 2021-07-02
Fix configure error with podman: permission for mountimg volume

Endle:
I was trying to compile proton 6.3 on Fedora 34 following https://github.com/ValveSoftware/Proton/#readme

```
mkdir build && cd build
../proton/configure.sh --container-engine=podman --enable-ccache
make dist
```

However, `configure.sh` had been complaining about no permission to read/write `/test`  

I think this issue is caused by SELinux, reference: https://www.redhat.com/sysadmin/user-namespaces-selinux-rootless-containers

> It blows up with permission denied. The user reads the man page, and figures out the problem is SELinux. The user sees that they can add a :Z option to the volume mount, which tells Podman to relabel the volume's content to match the label inside the container. And the SELinux problem is solved.

With this PR, I can start compling proton on my Fedora notebook.

I wonder, is it proper to add `:Z` to all volume mountings.  How about adding a new variable `$CONTAINER_VOLUME_MOUNTING_FLAG`, keep it empty by default. If SELinux is enabled, we set this variable to `:Z`

I had created a PR to compile proton 5.13 : https://github.com/ValveSoftware/Proton/pull/4634
I think proton 6.3 has a really good support to be compiled without the heavy vagrant. Is there still any interest to fix failed builds for 5.13?

Thank you

### DIFF
--- a/build/makefile_base.mak
+++ b/build/makefile_base.mak
@@ -58,7 +58,7 @@ CCACHE_ENV := $(patsubst %,-e %,$(shell env|cut -d= -f1|grep '^CCACHE_'))
 ifeq ($(ENABLE_CCACHE),1)
 	CCACHE_BIN := ccache
 	export CCACHE_DIR := $(if $(CCACHE_DIR),$(CCACHE_DIR),$(HOME)/.ccache)
-	DOCKER_OPTS := -v $(CCACHE_DIR):$(CCACHE_DIR) $(CCACHE_ENV) -e CCACHE_DIR=$(CCACHE_DIR) $(DOCKER_OPTS)
+	DOCKER_OPTS := -v $(CCACHE_DIR):$(CCACHE_DIR):Z $(CCACHE_ENV) -e CCACHE_DIR=$(CCACHE_DIR) $(DOCKER_OPTS)
 else
 	export CCACHE_DISABLE := 1
 	DOCKER_OPTS := $(CCACHE_ENV) -e CCACHE_DISABLE=1 $(DOCKER_OPTS)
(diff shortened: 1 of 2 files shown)


ivyl:
Hey, thanks for the report! That's a good catch, I don't use any SELinux-enabled distro myself.

Looks like docker also supports :Z for re-labeling, which is a good news as we won't need any special handling here. We can probably expose it as a configure flag and mention it in both `--help` and README, as "relabeling" sounds serious enough to be something that users opt-in for.

I'll have to read a bit more on what are the implications of relabeling and make sure it doesn't mess up anything outside of the container. I have to finish few other things though, but I should come back to you with the patches early next week.

As of 5.13 - AFAIU it is not expected to see any development, so I don't think there's any benefit in bringing the build system changes over there

Endle:
Hi,

I'm glad to hear the reply from you, and I'm glad to make changes to this PR if we can find an elegant way

> We can probably expose it as a configure flag and mention it in both --help and README  

I like this idea.

>  so I don't think there's any benefit in bringing the build system changes over there  

Me too. I've closed my PR #4634 

Thank you.

ivyl:
I've did a bit of reading, and having to opt-in for relabeling seems like the best option. This can be potentially dangerous when done on a wrong directory, like /usr or /etc. I don't expect that to happen with proton, but it's better to let know user what's happening.

I came up with this: https://github.com/ivyl/Proton/commit/2262af0da9cf92931191fd37e4135cccb7742eee

What do you think? Can you give it a spin?

Endle:
Hi,

I tested https://github.com/ivyl/Proton/commit/2262af0da9cf92931191fd37e4135cccb7742eee

```
git checkout ivyl/relabel-volumes
../proton/configure.sh --container-engine=podman --enable-ccache --relabel-volumes
```
The configuration step works perfectly now
Thank you

Endle:
Your patch looks quite clean. I like your solution :)

ivyl:
https://github.com/ivyl/Proton/commit/d071bcea72d23005f214e892017e9a23a86a045b will be in the next Proton release and experimental rebased on top of that. I think we can close this pull request.

Thanks again for the feedback.

Endle:
Thank you!

PR #5069, opened 2021-08-13
Proton isn't in beta version of steam client anymore

Weirdo1312:


aeikum:
Hey, thanks for the suggestion. It is still the case that sometimes new features and fixes get enabled in the Steam Beta Client first, so I think we should leave the note as it is.

arrowgent:
as how software goes, wine/proton is constantly improving.

its also fairly stable for many titles.

however, i think only proton-experimental is considered a "beta" release

while i understand that SteamPlay Beta Proton-Beta releases are pre-releases prior to being stable.

Ruedii:
I don't think Steam Beta is running separate Steam Runtime anymore.  

So, it might be good to remove most of this.

Probably replace this with:
"*Please note that Proton compatibility is still considered a Beta feature of Steam and may have noticeable bugs and limitations.  Occasionally certain features may initially require the Beta version of Steam."

ivyl:
Hi. This section was removed in the recent README rewrite. It's already in experimental and will be pushed to 6.3 upon next release.

PR #5130, opened 2021-08-31
[issues forms] new Forms design using YAML schema markdown, replacing templates

arrowgent:
i converted these from your existing markdown templates into the new YAML forms

example results:
https://github.com/arrowgent/Proton/issues
https://github.com/arrowgent/Proton/issues/1

_test yourself by starting an issue_

files changed:
https://github.com/arrowgent/Proton/tree/proton_6.3/.github/ISSUE_TEMPLATE

**recent Beta release:**
https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/

converting from template markdown to YAML:
https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

im new to YAML
i think `Version` can be reworked to automatically pull from Releases, if possible somehow to automatically add new lines here from your workflows?
so you dont have to always update the version in the report.yaml


aeikum:
This seems like a lot of overhead for not a whole lot of benefit. @kisak-valve I feel the current templates are working well from a developer perspective. Are users having issues filling out the current templates?

kisak-valve:
From my perspective, without this change, most of the feedback coming in is already structured well enough to be usable by Proton devs. This merge request would need substantial future-proofing adjustments so that Proton devs wouldn't need to change it on every major release. Also, in the interest of uniformity, it asks for more details in places we don't particularly care about.

Another downside from this conversion, with the current markdown-based templates, it's very easy to grab the template and use it elsewhere like as a comment of an existing issue report because it's closer to a human-readable format.

That said, overall, I don't have a strong opinion on the intent of this merge request.

arrowgent:
the version section is what i see needing more upkeep,
it could be changed to an input box asking the user to put the proton version
it does seem redundant since another area asks the same question


i agree many users do their best to translate data into the box for the issue `template` you already have
this just breaks it up into boxes and requirements (checkbox) giving a cleaner look

i just converted what _your current template_ said and put into the new `Forms` style
anything you'd like to change i can work on it (sepecifically the version selection)
or we can call it closed for a future idea

arrowgent:
closing since no interest.

PR #5157, opened 2021-09-14
proton: Delete PID leading zeros instead of all zeros

SSYSS000:
Currently, `gdb_attach` script deletes ALL the zeros of a process ID reported by `winedbg`, possibly under a false assumption that there are zeros only at the beginning. However, the PID can be reported as 0000010c or 000000c0 or whatever. This pull request fixes the issue by removing only the leading zeros.

### DIFF
--- a/proton
+++ b/proton
@@ -1169,7 +1169,7 @@ class Session:
             f.write("#Run winedbg in gdb mode and auto-attach to already-running program\n\n")
             f.write("cd \"" + os.getcwd() + "\"\n")
             f.write("EXE_NAME=${1:-\"" + exe_name + "\"}\n")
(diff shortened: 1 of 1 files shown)


aeikum:
@SSYSS000 Looks good to me. Thanks for the PR, will merge.

ivyl:
This has been pushed and is included in today's experimental https://github.com/ValveSoftware/Proton/commit/4a8ea09970b8eb8040fea07e1cdd3d73c6669a53

Jakeshadow:
https://steamcompatible.com/anti-cheat/ tracks compatibility for Deathloop and the broader list. Deathloop does work on SteamOS with the right Proton version, but performance varies a lot by hardware.

PR #5326, opened 2021-11-17
Set VKD3D_SHADER_DEBUG=none by default.

DadSchoorse:


ivyl:
Default level is fixme, right? Do you know how much logs does this produce with warn?

I am just curious and want to make sure that this won't render `PROTON_LOG=1` unusable.

DadSchoorse:
Right, I just matched `VKD3D_DEBUG` but it turns out `VKD3D_SHADER_DEBUG=warn` spams a lot.

ivyl:
I've tested it locally and indeed `warn` spams a lot. Way more than +seh or the other vkd3d-proton's channel.

I've pushed your change that uses the `fixme` level to experimental. We can fine-tune it later. Thanks.

PR #5375, opened 2021-12-03
Fix proton version on compat data files

marcelopio:
Current proton version is 6.3-8, but since CURRENT_PREFIX_VERSION is hardcoded to 6.3-3 it was displaying the wrong version on compatdata version files.

### DIFF
--- a/proton
+++ b/proton
@@ -30,7 +30,7 @@ from random import randrange
 #To enable debug logging, copy "user_settings.sample.py" to "user_settings.py"
 #and edit it if needed.
 
-CURRENT_PREFIX_VERSION="6.3-3"
+CURRENT_PREFIX_VERSION="6.3-8"
 
 PFX="Proton: "
 ld_path_var = "LD_LIBRARY_PATH"


marcelopio:
Although I made this fix, I think it should be automated some way, maybe by getting from dist/version?

ivyl:
It's bumped manually when there's something in the prefix that needs special fixups. See [`upgrade_pfx()`](https://github.com/ValveSoftware/Proton/blob/proton_6.3/proton#L373).

marcelopio:
I can understand that, but it is really frustrating when you are trying to troubleshoot why a game said to work on a specific version (Magicka - 6.3-8) does not, and then you go to the prefix and there is a file `version` that says `6.3-3`, and you look for hours why your prefix is not updating to `6.3-8` to just discover that it was the correct version all along, it was just the wrong input.

Is there any con of getting the current version from `dist/version`? 
It would fix the version mismatch and also continue to work for the fixups, since they check if the version is greater than needed.

kisak-valve:
Hello @marcelopio, previously noted at https://github.com/ValveSoftware/Proton/pull/4829#issuecomment-841861425, Proton version and wine prefix version are not the same thing. If there are no on-disk changes between patch revisions, then the wine prefix version is not bumped.

aeikum:
Yes, the plan is to change this to a letter after the next Wine merge to avoid future frustrations like this.

aeikum:
Turns out it's better to keep this value as an integer, so our plan is to use numbers starting at "100" for the prefix version. So the message will be something like, "updating prefix to version 7.0-100" which should hopefully make it obvious that it is not related to the proton version :)

PR #5414, opened 2021-12-16
steam_helper: Fix typo in json.h

eltociear:
assignement -> assignment


aeikum:
Hi, thanks for your contribution. This file was imported from an external project: https://github.com/open-source-parsers/jsoncpp

The right thing to do is fix the problem in the upstream project, and then update our copy of that project.

PR #5687, opened 2022-03-16
proton: Set DXVK_ENABLE_NVAPI envvar when NvAPI is enabled

liam-middlebrook:
See: https://github.com/doitsujin/dxvk/pull/2544 (merged)


ivyl:
pushed to our internal experimental branch, this will be included in the next automated bleeding-edge update and then in the next released experimental-proper, thanks!

PR #5688, opened 2022-03-16
Patch configure.sh for --no-proton-sdk

JustArchi:
Previously, `--no-proton-sdk` would effectively crash the configure step, initially on detecting the container system to use, then on undefined `ROOTLESS_CONTAINER` variable. I've patched it to skip the first, and properly set the variable based on the host system.

It allowed the configure step to succeed and started the host build for me, so I'd guess that it did the trick, but I'm not anywhere close to being proton expert, let alone knowing the source structure, so feel free to correct it if deemed appropriate.

Thank you in advance for considering this PR.

### DIFF
--- a/configure.sh
+++ b/configure.sh
@@ -141,20 +141,26 @@ function configure() {
     CONTAINER_MOUNT_OPTS=:Z
   fi
 
-  if [[ -n "$arg_container_engine" ]]; then
-    check_container_engine "$arg_container_engine" || die "Specified container engine \"$arg_container_engine\" doesn't work"
-  else
-    stat "Trying to find usable container engine."
-    if check_container_engine docker; then
-      arg_container_engine="docker"
-    elif check_container_engine podman; then
-      arg_container_engine="podman"
+  if [[ -n "$steamrt_image" ]]; then
+    if [[ -n "$arg_container_engine" ]]; then
+      check_container_engine "$arg_container_engine" || die "Specified container engine \"$arg_container_engine\" doesn't work"
     else
-        die "${arg_container_engine:-Container engine discovery} has failed. Please fix your setup."
+      stat "Trying to find usable container engine."
+      if check_container_engine docker; then
+        arg_container_engine="docker"
+      elif check_container_engine podman; then
+        arg_container_engine="podman"
+      else
+          die "${arg_container_engine:-Container engine discovery} has failed. Please fix your setup."
+      fi
     fi
-  fi
 
-  stat "Using $arg_container_engine."
+    stat "Using $arg_container_engine."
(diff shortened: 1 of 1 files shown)


aeikum:
@ivyl This one is for you to review, I think.

configure.sh
```diff
@@ -141,20 +141,26 @@ function configure() {
     CONTAINER_MOUNT_OPTS=:Z
   fi
 
-  if [[ -n "$arg_container_engine" ]]; then
```
ivyl, line 144: this can be made without introducing another level of nesting, e.g.:
```bash
if [[ -z "$steamrt_image" ]]; then
    stat "Using bare-metal build."
   # whatever special handling is necessary
elif [[ -n "$arg_container_engine" ]]; then
```

configure.sh
```diff
@@ -141,20 +141,26 @@ function configure() {
...
+        arg_container_engine="podman"
+      else
+          die "${arg_container_engine:-Container engine discovery} has failed. Please fix your setup."
+      fi
     fi
-  fi
 
-  stat "Using $arg_container_engine."
+    stat "Using $arg_container_engine."
+  elif [[ "$(id -u)" -eq 0 ]]; then
```
ivyl, line 159: I think if we are building on bare metal `ROOTLESS_CONTIANER` can always be `0`. It should not do anything meaningful without podman/docker anyway. You can just set in in the  `if [[ -z "$steamrt_image" ]]; then` section.

This check should be not necessary and people generally should not build as a root :-)

ivyl (commented):
Sorry for late review. This has slipped through cracks.


The sanity check logic also needs some tweaking:
```bash
# Sanity check arguments
if [[ -n $arg_no_protonsdk && -n $arg_protonsdk_image ]]; then
    die "Cannot specify --proton-sdk-image as well as --no-proton-sdk"
elif [[ -z $arg_no_protonsdk && -z $arg_protonsdk_image ]]; then
    die "Must specify either --no-proton-sdk or --proton-sdk-image"
fi
```
with the current logic `../configure.sh --no-proton-sdk` complains about `!! Cannot specify --proton-sdk-image as well as --no-proton-sdk` because `$arg_proton_sdk_image` has a default value nowadays.

JustArchi:
To be honest I don't even know if it still makes sense. I tried to build proton with `--no-proton-sdk` but failed regardless at some later stage, likely due to missing dependencies (which was to be expected).

I don't know if anybody finds `--no-proton-sdk` a valid use case even. I think it'll be better and much more efficient for me to just close this PR as I don't find it justified anymore, and if you believe it should be allowed, you'll make the required fixes far faster than me anyway.

Thank you a lot for your work on Proton @ivyl, I appreciate it a lot.

PR #5891, opened 2022-06-08
Add support for forwarding commands into the container environment

smcv:
Developed against ~~`proton_7.0-rc`~~ Proton Experimental.

This version requires https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/465, and doesn't alter the `PROTON_DUMP_DEBUG_COMMANDS` scripts. Old version at https://github.com/smcv/Proton/tree/launcher-service1 for reference.

To test with current versions of `SteamLinuxRuntime_soldier` and Steam, some reordering of chickens and eggs is required:

* The Steam client doesn't yet know how to activate command-injection into compat tools, so set `STEAM_COMPAT_LAUNCHER_SERVICE=proton` in Steam's environment or in a game's Launch Options. This will stop being necessary when the work by @TTimo that this is intended to enable gets a bit further.
* ~~SLR_soldier doesn't have the new `steam-runtime-launcher-interface-0`, so replace `SteamLinuxRuntime_soldier/pressure-vessel` with a build from https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/merge_requests/465. https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/jobs/95761/artifacts/raw/_build/production/pressure-vessel-bin.tar.gz should be suitable. This will stop being necessary when `launcher-interface-0` gets merged and released.~~ SLR_soldier `client_beta` is suitable.
* ~~The new `steam-runtime-launcher-interface-0` isn't in `PATH` yet, so set `PRESSURE_VESSEL_SHELL=instead` in Steam's environment or in a game's Launch Options. When you start the game, instead of the actual game, you'll get an `xterm` running in the container. In that `xterm`, use `export PATH=/run/pressure-vessel/pv-from-host/bin:$PATH` to get the necessary tool into the `PATH`, then run `"$@"` (including the quotes) to run Proton and the actual game. This will stop being necessary when `launcher-interface-0` gets merged and released.~~ SLR_soldier `client_beta` is suitable.

~~https://gitlab.steamos.cloud/steamrt/steamlinuxruntime/-/merge_requests/76 is the corresponding glue to get this launched in the soldier/sniper container (`STEAM_COMPAT_LAUNCHER_SERVICE=container-runtime`) or in the scout-on-soldier "layered" environment that can be used for native Linux games (`STEAM_COMPAT_LAUNCHER_SERVICE=scout-in-container`).~~ (This has been integrated.)

---

* proton: Allow forwarding commands into the Proton environment
    
    Recent versions of the Steam Runtime include an IPC server/client pair
    which can be used to run commands inside the container environment
    (or any other special execution environment), analogous to sshd/ssh or
    flatpak-portal/flatpak-spawn. The server runs inside the Steam Runtime
    container and accepts commands over D-Bus; the client runs on the host
    system, asks the server to run a command, and forwards its stdin, stdout
    and stderr back to the host.
    
    https://gitlab.steamos.cloud/steamrt/steamlinuxruntime/-/merge_requests/72
    adds support for injecting commands into the SteamLinuxRuntime_soldier
    compatibility tool (and any later version, such as sniper). However,
    Steam compatibility tools are stackable: in particular, Proton runs in a
    soldier container (or presumably sniper in future). If we are debugging
    a Proton game, then ideally we will want to inject commands into Proton's
    execution environment rather than soldier's, so that they run with the
    correct environment variables etc. to communicate with a running Proton
    session. In particular, it's important that the `WINEPREFIX` is correct.
    
    The steam-runtime-launcher-interface-0 program implements the
    interface for compatibility tools to use to decide where, if anywhere,
    to launch the command server.
    
    This commit does not alter the scripts produced by
    PROTON_DUMP_DEBUG_COMMANDS. To run those scripts' commands in the
    container environment, pass their filenames to
    steam-runtime-launch-client.

### DIFF
--- a/proton
+++ b/proton
@@ -1450,6 +1450,11 @@ class Session:
         return subprocess.call(args, env=local_env, stderr=self.log_file, stdout=self.log_file)
 
     def run(self):
+        if shutil.which('steam-runtime-launcher-interface-0') is not None:
+            adverb = ['steam-runtime-launcher-interface-0', 'proton']
+        else:
+            adverb = []
+
         if "PROTON_DUMP_DEBUG_COMMANDS" in self.env and nonzero(self.env["PROTON_DUMP_DEBUG_COMMANDS"]):
             try:
                 self.dump_dbg_scripts()
@@ -1467,9 +1472,11 @@ class Session:
 
         # CoD: Black Ops 3 workaround
         if os.environ.get("SteamGameId", 0) == "311210":
-            rc = self.run_proc([g_proton.wine_bin, "c:\\Program Files (x86)\\Steam\\steam.exe"] + sys.argv[2:] + self.cmdlineappend)
+            argv = [g_proton.wine_bin, "c:\\Program Files (x86)\\Steam\\steam.exe"]
         else:
-            rc = self.run_proc([g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"] + sys.argv[2:] + self.cmdlineappend)
+            argv = [g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"]
+
+        rc = self.run_proc(adverb + argv + sys.argv[2:] + self.cmdlineappend)
 
         if remote_debug_proc:
             remote_debug_proc.kill()
--- a/toolmanifest_runtime.vdf
+++ b/toolmanifest_runtime.vdf
@@ -4,4 +4,5 @@
   "commandline" "/proton %verb%"
   "require_tool_appid" "1391110"
   "use_sessions" "1"
+  "compatmanager_layer_name" "proton"
 }


proton
```diff
@@ -28,10 +28,18 @@ from ctypes import c_void_p
...
+    from shlex import quote as shell_quote
+except ImportError:
+    # exists in older versions, undocumented
+    from pipes import quote as shell_quote
+
 #To enable debug logging, copy "user_settings.sample.py" to "user_settings.py"
 #and edit it if needed.
 
 CURRENT_PREFIX_VERSION="7.0-100"
+PROTON_APP_ID="1887720"
```
smcv: I don't know whether you'll want to have some automation here to set the correct app-ID for each branch at build time, or just hard-code it in each branch and change it when merging experimental into a production branch, or what. For now I've kept it simple.

In the `SteamLinuxRuntime` branch, the startup scripts also read a `steam_appid.txt` file from the compatibility tool's top directory if present, analogous to the way the Steamworks API reads that file from the game's top directory when a game wants to figure out which Steam game it is. I haven't implemented that here, but it should be easy.

proton
```diff
@@ -1267,16 +1284,40 @@ class Session:
...
-    def dump_dbg_scripts(self):
+    def get_steam_runtime_launch_client_command(self, launcher_address):
+        launch_client = os.path.expanduser("~/.steam/root/ubuntu12_32/steam-runtime/amd64/bin/steam-runtime-launch-client")
+
+        # This relies on an implementation detail of
+        # SteamLinuxRuntime_soldier, but is necessary until the scout
+        # runtime shipped with the GA Steam client has been updated to
+        # include a copy of s-r-launch-client.
+        if "PRESSURE_VESSEL_RUNTIME_BASE" in self.env and not os.path.exists(launch_client):
+            launch_client = self.env["PRESSURE_VESSEL_RUNTIME_BASE"] + "/pressure-vessel/bin/steam-runtime-launch-client"
```
smcv: I expect this part to become unnecessary when the next Steam beta cycle gets promoted to GA.

proton
```diff
@@ -1267,16 +1284,40 @@ class Session:
...
+        if launcher_address:
+            launch_client = self.get_steam_runtime_launch_client_command(launcher_address) + "\tenv \\\n"
+        else:
+            launch_client = ""
+
         with open(tmpdir + "winedbg", "w") as f:
             f.write("#!/bin/bash\n")
             f.write("#Run winedbg with args\n\n")
             f.write("cd \"" + os.getcwd() + "\"\n")
+            # TODO: Should this run via launch_client?
```
smcv: I wasn't sure how `winedbg` works in container-world: is it something that we want to run inside the container, or on the host system? If we want to run it inside the container, then this should get the same treatment as `run`.

proton
```diff
@@ -1294,6 +1335,7 @@ class Session:
                 else:
                     f.write(" \"" + arg + "\"")
             f.write(")\n")
+            # TODO: Should this run via launch_client?
```
smcv: Same as `winedbg` above.

proton
```diff
@@ -1343,7 +1385,8 @@ class Session:
                 else:
                     f.write(" \"" + arg + "\"")
             f.write(")\n")
-            self.dump_dbg_env(f)
+            f.write(launch_client)
+            self.dump_dbg_env(f, using_launcher_service=bool(launcher_address))
             f.write("\t\"" + g_proton.wine64_bin + "\" c:\\\\windows\\\\system32\\\\steam.exe \"${@:-${DEF_CMD[@]}}\"\n")
```
smcv: This seems to basically work, but I'm not sure whether I'm using it right.

With this `./proton` replacing the one from Proton 7.0-2, I can run

```
.../SteamLinuxRuntime_soldier/pressure-vessel/bin/steam-runtime-launch-client \
        --bus-name=com.steampowered.App312990 \
        --directory='' \
        -- \
        xterm
```

and I'll get an `xterm` in the Proton environment (`WINEPREFIX`, etc. are set). App ID 312990 is Expendabros (a promo/demo version of Broforce), which is one of my usual test games if I don't particularly care what Windows game I'm running.

I can also run commands like

```
/tmp/proton_$(id -nu)/run cmd
/tmp/proton_$(id -nu)/run winecfg
```

but they don't exit when `cmd` or `winecfg` does. It looks as though when Proton runs `wine64 c:\windows\system32\steam.exe cmd`, the `cmd` process correctly exits when closed, but the `steam.exe` process keeps running until I send `SIGTERM` to it. What's the intention of wrapping the process in `steam.exe` here?

Workaround: if you kill the `steam-runtime-launch-client` process with `SIGTERM` (for example Ctrl+Z, `bg`, `kill %%`), it will forward that signal to `steam.exe`, which will exit.

smcv: I'm not sure that I'm understanding correctly what the purpose of `/tmp/proton_$(id -nu)/run` is, because it defaults to running the actual (Windows) game; but in the container environment, when the "main" copy of the Windows game exits (whether successfully or by crashing), we'll tear down the container, which will terminate `steam-runtime-launcher-service`, and this script will no longer be able to poke commands into it?

If we had a mode where `steam-runtime-launcher-service` can be told to keep running after the command it wraps has exited, which is something I plan to implement at some point, then that would make this script more useful.

proton
```diff
@@ -1370,9 +1427,11 @@ class Session:
 
         # CoD: Black Ops 3 workaround
         if os.environ.get("SteamGameId", 0) == "311210":
-            rc = self.run_proc([g_proton.wine_bin, "c:\\Program Files (x86)\\Steam\\steam.exe"] + sys.argv[2:] + self.cmdlineappend)
+            argv = [g_proton.wine_bin, "c:\\Program Files (x86)\\Steam\\steam.exe"]
         else:
-            rc = self.run_proc([g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"] + sys.argv[2:] + self.cmdlineappend)
+            argv = [g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"]
+
+        rc = self.run_proc(adverb + argv + sys.argv[2:] + self.cmdlineappend)
```
smcv, line 1479: Not really related to this feature at all, but if we *don't* have a remote debug process, it might be nicer if the `proton` script did an `execve()` to completely replace itself with the `wine64 ...` command?

ivyl:
I don't think I fully understand how we ended up with this design. Why can't one just execute `proton` from inside the SRT environment that has all the variables set already? `proton` infers things like `WINEPREFIX` from `STEAM_COMPAT_DATA_PATH` anyway.

Also I'm not sure if I understand the intent here correctly or how it makes this easier. Correct me if I'm wrong but the intended use case seems to be:

1. Start a game with the magic environment variable `STEAM_COMPAT_LAUNCHER_SERVICE=proton` whether that's done via Steam / Steam OS development kit / launch options.
2. `proton` then starts `wine steam.exe game.exe` via `steam-runtime-launcher-service` instead of doing that directly
3. it then prints a message on stderr with a magic command that allows you to execute more stuff with the same environment variables and in the same SRT session that was used for the `wine ...` execution

`wine` is not in `PATH` so I don't see what value does this provide over having just the SRT steam-runtime-launcher-service and invoking `proton` from there. It could be even set as `$PROTON`.

We may want to add a view more verbs to `proton` like gdb, ps, etc. which IMO would be more helpful than having the enviroment variables alone.

proton
ivyl: how is this supposed to work with tools that are dropped in `compatibilitytools.d/`? Also proton being alias to `1887720` feels wrong. Will you point it to Proton 8 (or whatever version the next rabase will be based on) when it releases? Maybe it should be just proton7?

proton
```diff
@@ -1352,10 +1395,24 @@ class Session:
...
+            appid = os.environ.get("STEAM_COMPAT_APP_ID", os.environ.get("SteamAppId", "0"))
+            launcher_address = "--bus-name=com.steampowered.App" + appid
+            adverb = ['steam-runtime-launcher-service', '--replace', launcher_address, '--']
+
+            sys.stderr.write("Starting program with command-launcher service.\n")
+            sys.stderr.write("\n")
+            sys.stderr.write("To inject commands into the container, use a command like:\n")
+            sys.stderr.write("\n")
+            sys.stderr.write(self.get_steam_runtime_launch_client_command(launcher_address))
+            sys.stderr.write("\txterm\n")
```
ivyl: Why is this printed to stderr? How is supposed to see that message? Why isn't hat just a part of Proton's log?

proton
```diff
@@ -28,10 +28,18 @@ from ctypes import c_void_p
 from filelock import FileLock
 from random import randrange
 
+try:
+    # available since 3.3
+    from shlex import quote as shell_quote
```
ivyl: I think it should be okay to hard require something newer than 3.3. We are already using f-strings.

proton
ivyl: `winedbg` needs to be started inside of the prefix. It's an `.exe` and works with processes on the Windows side of things.

proton
ivyl: The intention of those script is to run the game directly from the very start without the container. It's mostly for debugging purposes so you can see the full environment and make changes to what's started.

I think most people nowadays just attach gdb running outside of the container to the running game anyway + maybe `PROTON_WAIT_ATTACH`.

smcv:
> Why can't one just execute proton from inside the SRT environment that has all the variables set already?

It's a container environment (analogous to Flatpak or Docker or whatever), so setting environment variables isn't sufficient to "enter" it. To see the same libraries that Proton depends on, and more generally the same view of the filesystem that Proton has, you can't just run commands on the host system. Instead, we need an IPC protocol with a server inside the SRT environment, and a client outside that sends commands to the server: analogous to having an `sshd` inside the SRT environment and connecting to it with a `ssh` client on the outside, or using `docker exec` to poke commands into a running Docker container.

Because this is all local to one machine, we don't need the encryption or networking that `sshd` would provide, and we can use D-Bus over `AF_UNIX` sockets instead. `steam-runtime-launcher-service` is the server side of that, and `steam-runtime-launch-client` is the client.

The alternative to using IPC would be to do complicated things with namespaces to force a process into the container from the outside. `flatpak enter` does that, and it's really horrible (sometimes needs root, sometimes fails in mysterious ways with unclear error messages, and you don't necessarily even get the same execution environment you were expecting), so eventually I want to replace `flatpak enter` with a similar IPC mechanism.

proton
smcv: It's a proof-of-concept, I don't have an in-depth understanding of how Proton fits together internally. If there's a better place to write this information, then it can certainly go there instead.

proton
smcv: I believe the intended design on the Steam side is that if a developer has requested that the game is run with the ability to inject arbitrary debugging commands like this, then the Steam client will iterate through the stack of compatibility tools and identify whichever one is the last one before the game itself:

```
Steam client -> SteamLinuxRuntime_soldier -> Proton (whatever version) -> a Windows game
                                                      ^^^

Steam client -> SteamLinuxRuntime_soldier -> SteamLinuxRuntime -> an old native Linux game
                                                      ^^^

Steam client -> SteamLinuxRuntime_sniper -> a future native Linux game
                         ^^^
```

and then set `STEAM_COMPAT_LAUNCHER_SERVICE` to the app-ID of that tool, so that the arbitrary commands get injected into something that is as close as possible to the execution environment of the actual game.

If compatibility tools have a unique identifier, then we could use that unique identifier instead of the app-ID (assuming @ttimo or his colleagues can implement that on the Steam side); but I don't have visibility into the `compatibilitytool.vdf` of the various SteamLinuxRuntime tools, so I don't know what their unique identifiers are. I do know what their app IDs are, and their app IDs are how Proton's `toolmanifest.vdf` refers to them in any case, so that seemed a reasonable unique identifier to work with.

Strictly speaking, it only needs to be unique within a "stack" of compatibility tools: tools A and B only need to be distinguished if it would make sense to run tool A "inside" tool B or vice versa. But it needs to be something that the Steam client knows, so using an existing metadata field seemed easier than inventing a new one.

> how is this supposed to work with tools that are dropped in `compatibilitytools.d/`?

It won't, because Steam can't know an app-ID for those. But if the developer sets `STEAM_COMPAT_LAUNCHER_SERVICE=proton`, then Steam will ignore it and pass it through, and then the startup script will recognise that as a request to have Proton be the layer that starts the `launcher-service`.

> Also proton being alias to 1887720 feels wrong. Will you point it to Proton 8 (or whatever version the next rabase will be based on) when it releases? Maybe it should be just proton7?

There's no need to distinguish between `proton7` and `proton8` unless there would be a legitimate scenario in which the stack of compatibility tools includes more than one version of Proton (Proton running inside Proton), which I don't think would ever make sense? So the alias that's there for developer convenience is just `proton`, meaning "whatever stack of compatibility tools I'm using, if it includes some version of Proton, then I want the `launcher-service` to be inserted at that layer".

The app-ID here needs to match up with what the Steam client will set as the value of `STEAM_COMPAT_LAUNCHER_SERVICE`, which will be different for each Proton app-ID in the Steam catalogue. Proton has separate git branches for 5.13, 6.3, 7.0 and experimental, so I had assumed that it would be easy to change the app-ID in `./proton` when switching between git branches.

Alternatively, if it needs to be possible to set an app-ID "from the outside", then it could be injected via the build system instead, similar to the way the `$(build_name)` is handled: that's what I did for `SteamLinuxRuntime_soldier` and `SteamLinuxRuntime_sniper`, which are built from the same source code and just have different parameters passed in.

proton
smcv: > without the container

So is that assuming that your host library stack is "close enough" to soldier that you can run Proton/Wine binaries built on soldier using your host library stack?

If that's the case, then this one should *not* be wrapped in `steam-runtime-launch-client`, but `winedbg` probably should.

proton
smcv: The command is always the same (on a per-game-ID basis) so if you are regularly debugging a particular game, you can just keep using the same command.

smcv:
> Why can't one just execute proton from inside the SRT environment that has all the variables set already?

The Steam container runtime doesn't know anything about Proton: it's just a generic library stack, and doesn't know whether the command it is going to be running is Proton or a native Linux game. It's only the Proton scripts that know how Proton/Wine works, and therefore how to get a command running in the desired execution environment.

> wine is not in PATH

It could be if you wanted it to be, but it would have to be the Proton startup script that did this, because the Steam container runtime does not know anything about Proton.

> It could be even set as $PROTON.

Again, the container runtime is not in a position to do this, because it has no idea what its "payload" is: it could be a native Linux game, it could be Proton, it could be dosbox, ...

proton
smcv: Ah, I hadn't found those. They're new syntax in Python 3.6, so Proton requires at least 3.6, therefore I can use `shlex.quote` unconditionally.

smcv:
> We may want to add a [few] more verbs to proton like gdb, ps, etc. which IMO would be more helpful than having the enviroment variables alone.

Part of the idea of this command-injection interface was that the compatibility tool developer doesn't have to anticipate everything that the game developer could possibly want: if the debugging entry point is "run an arbitrary command", then that command can be anything that exists in the container. Things like `gdb` and `strace` won't be available unless you've modified the container to use a SDK runtime instead of the usual Platform runtime, but that's something I want to streamline at some point (at the moment it's possible-but-annoying). The normal Platform runtime does have `gdbserver`, which is enough to attach a full-fat `gdb` on the host.

For the diagnostic tools that are included in Wine/Proton, the container runtime has no idea what Proton is or where to find it, but Proton does know that, and is in a position to add `wine`, `winedbg`, etc. to the `PATH` if it wants to, or set an environment variable from which a developer can find them. If you want a way to get particular variables to be set for the debugging subprocesses started by `steam-runtime-launcher-service`, but without affecting the environment of the "main process" (`steam.exe` and/or the actual game), that's not *currently* possible, but I could add an `--env=FOO=bar` option? For instance I could imagine that you might want to add the Wine executables to the `PATH` that is presented to the debugging commands, but without altering the `PATH` of the actual game.

For debugging Windows games specifically, it seems like there are two things you could reasonably want to do: one is to run some Linux program like `gdbserver`, and the other is to (use `wine` or `wine64` to) run some Windows program that will interact with your `WINEPREFIX` as though it was running in native Windows, for instance Wine's `cmd` or `regedit`, or a Windows/MSVC equivalent of `gdbserver` or `strace`.

TTimo:
@ivyl - sorry this probably comes very out of context.

The internal task is https://gitlab.steamos.cloud/jupiter/tasks/-/issues/573, which CW should have access to. @nsivov will have more context from the initial email thread with Valve.

This particular patch is from the prerequisite work in https://gitlab.steamos.cloud/jupiter/tasks/-/issues/483.

I have been able to use an earlier version of this patch to trigger the `getminidump.exe` that @nsivov wrote, with a command like this:

`/home/timo/steam/main/client/ubuntu12_32/steam-runtime/amd64/usr/bin/steam-runtime-launch-client --bus-name=com.steampowered.App480 -- "/opt/steam/steamapps/common/Proton - Experimental/files/bin/wine64" /home/timo/Downloads/getminidump.exe 72414`

Having the launcher service below the proton script in the hierarchy is working well. As Simon pointed out, additional verbs would impair our ability to easily extend the functionality, and also requires ad-hoc knowledge in the Steam client which I'm trying to avoid for the same extensibility reasons.

ivyl:
@smcv

> It's a container environment (analogous to Flatpak or Docker or whatever), so setting environment variables isn't sufficient to "enter" it. 

> The Steam container runtime doesn't know anything about Proton: it's just a generic library stack, and doesn't know whether the command it is going to be running is Proton or a native Linux game. It's only the Proton scripts that know how Proton/Wine works, and therefore how to get a command running in the desired execution environment.

I get this. The question is what's the benefit of `STEAM_COMPAT_LAUNCHER_SERVICE=proton` over `container-runtime`? Can't we just use the container-runtime and call `proton` script from there?

> > wine is not in PATH

> It could be if you wanted it to be, but it would have to be the Proton startup script that did this, because the Steam container runtime does not know anything about Proton.

True. If we go this way I think this is something we would need to add.

> Again, the container runtime is not in a position to do this, because it has no idea what its "payload" is: it could be a native Linux game, it could be Proton, it could be dosbox, ...

I see Steam as proton / compatdata manager and Proton mostly as a clueless runner. To me it feels like setting `$PROTON` would be Steam's job.

> Part of the idea of this command-injection interface was that the compatibility tool developer doesn't have to anticipate everything that the game developer could possibly want: if the debugging entry point is "run an arbitrary command", then that command can be anything that exists in the container. Things like gdb and strace won't be available unless you've modified the container to use a SDK runtime instead of the usual Platform runtime, but that's something I want to streamline at some point (at the moment it's possible-but-annoying). The normal Platform runtime does have gdbserver, which is enough to attach a full-fat gdb on the host.

Sounds like something that would be solved via `$PROTON exec-linux ...` or similar.

> For the diagnostic tools that are included in Wine/Proton, the container runtime has no idea what Proton is or where to find it, but Proton does know that, and is in a position to add wine, winedbg, etc. to the PATH if it wants to, or set an environment variable from which a developer can find them. If you want a way to get particular variables to be set for the debugging subprocesses started by steam-runtime-launcher-service, but without affecting the environment of the "main process" (steam.exe and/or the actual game), that's not currently possible, but I could add an --env=FOO=bar option? For instance I could imagine that you might want to add the Wine executables to the PATH that is presented to the debugging commands, but without altering the PATH of the actual game.

I think adding wine's bin/ to PATH won't hurt in general. I don't think it would mess with games if we would do that unconditionally.

> For debugging Windows games specifically, it seems like there are two things you could reasonably want to do: one is to run some Linux program like gdbserver, and the other is to (use wine or wine64 to) run some Windows program that will interact with your WINEPREFIX as though it was running in native Windows, for instance Wine's cmd or regedit, or a Windows/MSVC equivalent of gdbserver or strace.

It continues to eludes me why we want to add all that complexity and inscribe all the runtime implementation details into `proton`. Prefixing our `wine` invocation with `PROTON_WINE_ADVERB` set by something external (Steam?) feels less leaky.

I'm still trying to understand why `proton` having the awareness and calling `steam-runtime-launch-server` is this the preferred way.

@TTimo 

Thanks for the context. I believe that with `STEAM_COMPAT_LAUNCHER_SERVICE=container-runtime` the following command would do the same thing:
```
steam-runtime-launch-client --bus-name=com.steampowered.App480 -- "/opt/steam/steamapps/common/Proton - Experimental/proton" run /home/timo/Downloads/getminidump.exe 72414
```

> Having the launcher service below the proton script in the hierarchy is working well. As Simon pointed out, additional verbs would impair our ability to easily extend the functionality, and also requires ad-hoc knowledge in the Steam client which I'm trying to avoid for the same extensibility reasons.

I see. How about having a generic verb that just executes anything with Proton's environment? I'm just not a fan of putting all the knowledge about how the runtime works into `proton` script. I'm trying to make sure that we haven't forgot to consider simpler options first.

smcv:
> I see Steam as proton / compatdata manager and Proton mostly as a clueless runner. To me it feels like setting $PROTON would be Steam's job.

Steam doesn't really know what Proton is or what it does, though; it has a generic concept of compatibility tools, but it doesn't know the difference between Proton (which runs Windows executables on Linux), SLR (which runs Linux executables on Linux), or a hypothetical dosbox compatibility tool (which runs DOS executables on Linux). All Steam knows is that there is a compatibility tool that it has been configured to run the game under (in this case Proton), and that because that tool has a dependency in its `toolmanifest.vdf`, it needs to wrap an "outer" compatibility tool (in this case SLR_soldier) around it.

When we run native Linux games under `SteamLinuxRuntime`, Steam is doing exactly the same things - but instead of having a compatibility tool to run Windows games (this is Proton) with a dependency (this is SLR_soldier), it's using a compatibility tool to run Linux games built for the `scout` runtime (this is `SteamLinuxRuntime`) with a dependency (this is SLR_soldier again).

> Prefixing our wine invocation with PROTON_WINE_ADVERB set by something external (Steam?) feels less leaky.

That would require Steam to have domain-specific knowledge about Proton being special (or more generically, a configuration parameter that basically has the semantics of "this is Proton and is special"), which is something we're trying to avoid. I thought a lot of the purpose of Proton was that it encapsulates the domain-specific knowledge that is involved in using (a suitably modified version of) Wine as a Steam compat tool? (As opposed to what might go in a different compat tool that used Dosbox or ScummVM or something.)

> I'm just not a fan of putting all the knowledge about how the runtime works into proton script.

I'm treating the IPC server and the `STEAM_COMPAT_LAUNCHER_SERVER` mechanism as being part of the wider Steam runtime environment here, rather than being something specific to pressure-vessel or the container runtime. At the moment we're getting the actual executables from pressure-vessel as a way to bootstrap, because we're able to iterate versions of pressure-vessel quicker than versions of the entire Steam client (I don't have access to new Steam clients until the general public does), but they'll eventually become part of the infrastructure that ships as part of the Steam client. Proton can find the `launcher-server` via the `PATH` (implementation detail: it's included in soldier) or, eventually, accompanying Steam itself, while debugging clients should eventually be getting the `launch-client` from Steam itself.

If we were only dealing with native Linux processes, then I would have suggested that Steam should treat the `launcher-server` as being yet another "adverb" to stuff onto the command-line after the last compat tool, like this:

```
/path/to/reaper ... -- \
/path/to/SteamLinuxRuntime_soldier/_v2-entry-point --verb=waitforexitandrun -- \
/path/to/Proton/proton waitforexitandrun \
steam-runtime-launcher-server --bus-name=...App302380 -- \
"/path/to/Floating Point.exe"
```

But that doesn't work in the presence of Proton (or other compat tools that translate between incompatible calling conventions, like dosbox), because the calling convention for a Windows-onto-Linux compat tool like Proton is that it expects a path to a Windows executable as its parameter, not a path to a Linux executable.

I'm also reluctant to suggest having the "API" involve Steam setting something like `STEAM_COMPAT_TOOL_WRAPPER=1887720:steam-runtime-launcher-server --bus-name=...App302380 --`. That would require hardwiring appropriate options for the launcher server into Steam, which updates on a slower schedule than the compat tools; would require all compat tools to be able to parse arbitrary strings into `argv` and get it right; and would require an assumption that the various different compat tools all want the same options, which I don't think is necessarily going to be true (for instance if Proton wanted to put `wine` onto the `PATH` for debug commands, without affecting the `PATH` seen by the actual game, then that would be Proton-specific rather than something that is true universally).

> > Part of the idea of this command-injection interface was that the compatibility tool developer doesn't have to anticipate everything that the game developer could possibly want
>
> Sounds like something that would be solved via `$PROTON exec-linux ...` or similar.

Right, that would be better than having `gdb`, `strace`, etc. as verbs.

However, this would mean the `proton` script needs to reproduce the same execution environment as for the main program, to the best of its ability. Environment variables are a big part of execution environment, yes, but there is a lot of other inheritable state involved in a Linux process's execution environment: where are you in the process hierarchy? is there a subreaper? what are the rlimits? do you have a controlling terminal? are you in a cgroup? and so on. Some of this is not possible to jump into from outside, and has to be inherited, otherwise it just doesn't work.

If we run the IPC server inside the last compat tool in the chain (in this case Proton), then it is a child of the compat tool, so it automatically inherits all the inheritable state from the compat tool, including future kernel concepts that haven't even been invented yet. That seems like a good thing to me. The last one in the chain also makes sense to me as being the special case, because there is already something special about it: it's the last one.

Another interesting question for this approach would be: who runs Proton with that verb, how, and in what environment?

If it works the same as `proton waitforexitandrun` and friends, then the answer would be: take the stack of compat tools, iterate through them until you reach Proton, go back one, and execute it in that environment. But that environment is `SteamLinuxRuntime_soldier`, which knows literally nothing about Proton: all it knows is that it was told to start a container, then run an arbitrary command in the container. It doesn't even know which part of the arbitrary command is the path to the Proton script: currently the arguments it gets happen to be `/path/to/proton waitforexitandrun /path/to/foo.exe`, but there's no reason why Steam couldn't be telling it to run `xterm -e env --chdir=/ --block-signal=SIGINFO FOO=bar nice 10 steam-add-game-overlay gdbserver localhost:12345 /path/to/proton waitforexitandrun /path/to/foo.exe`, and you can't find the Proton script in there without domain-specific knowledge.

Meanwhile, it's not enough for Steam to run the `proton` command in *a* soldier container: to be useful, it needs to run it in *the same* soldier container as the game itself. We've been here before with the setup commands that some games have, to run `installscript.vdf` or otherwise do setup tasks before the actual game to start; at the beginning of the Proton 5.13 cycle, each of those commands ran in a new soldier container instance, which was slow and didn't work properly.

So now someone (Steam?) needs to know at what point to "cut" the chain of compat tools at one that supports the IPC server interface, use the IPC client/server pair to replace that point and everything before it, but then find the Proton script among the rest and run it in the environment that the IPC server provides?

TTimo:
I generally agree with Simon's comments above. In particular that neither the compat tool(s) above Proton, nor Steam itself have knowledge of what they are running and we'd like to keep it that way.

I also understand wanting to keep Proton generally agnostic to being run within a compat tool environment vs not (a local dev host build for instance). It would be cleaner.

Maybe one option is that the compat tool version has a different `compat-tool-proton` entry point, which would kick off the launcher service, and invoke `proton`. Then when we run commands the `proton` script can be executed again with a `run` verb.

I don't think this extra complexity is very justified though, and I'm concerned about re-executing `proton`, doing the same discovery as the title launch etc. I'd much rather see the Proton script evolve to cleanly understand when it's under a compat tool context.

ivyl:
Thank you both for detailed explanations. This has changed my view on the place of Proton in the full ecosystem. Indeed, if you aim for having more compatibility tools than just SLR and Proton it makes sense for Steam to not have detailed knowledge about them. Also `toolmanifest.vdf`, as you've pointed, describes Proton's dependency on SLR. I think we can find a way of integrating the use of launcher service that is a bit cleaner than this initial draft and in a way that won't impede the developers.

I have few general questions that still bog me:

1. What's the correct scope for the launcher service? I find it a bit weird that it's inside SLR but outside or rather in the middle of Proton. It seems inconsistent that on the Proton side we stay on the Linux side and do not go inside Windows. If the two would be equivalent we should be executing things on the Windows side of things.
2. Why does each compatibility tool have to decide to run via `steam-runtime-launch-server` on its own? Can we have some runner that would be used unconditionally, if present, and make the decision for itself if it sets up the server or just ends up doing execv?
3. There should really be a better way of getting your own appid than having it hardcoded. I can automate creating `steam_appid.txt` when using `steamcmd` to upload build to Steam, because that's the only point our build system / CI needs to know this, but this still feels hacky.

smcv:
> What's the correct scope for the launcher service? I find it a bit weird that it's inside SLR but outside or rather in the middle of Proton. It seems inconsistent that on the Proton side we stay on the Linux side and do not go inside Windows.

You're right that this is a bit odd, but the `s-r-launcher-service` is a Linux executable, using Linux mechanisms for IPC (D-Bus over `AF_UNIX`, with Unix-specific fd-passing to tunnel the stdin/stdout/stderr file descriptors from outside the container), so you can't run it as a Windows executable. (Or can you? I don't know what happens if a Windows executable being run via Wine tries to run a Linux executable by its path or by `PATH` search.)

I inserted it at the last point in the chain where we're still in Linux-world, just before handing over to Windows-world, so that it's as close to the actual game's execution environment as it can be.

An advantage of this is that when poking in commands from the outside, the interpretation is consistent: if you're currently running Steam game 440 (which happens to be TF2), you can always run `s-r-launch-client --bus-name=com.steampowered.App440 -- xterm` and get an interactive Linux shell as close as possible to the game itself, without needing to know whether the currently-active session for game 440 is TF2-for-Windows (in which case the shell is "inside" the `./proton` wrapper, but "outside" Wine and the game's own launcher if any) or TF2-for-Linux (in which case the shell is "inside" any SLR_soldier and optionally SLR scout environment, but "outside" the game's startup scripts). The design principle is that it's the point closest to the game at which (a) we're still in "Linux world", and (b) compat tools still have control over what happens (by the time we get into a native Linux game's startup scripts, it's too late to intervene).

If we were using a Windows executable as the IPC service, then it would be impossible to use the same implementation we use for Linux games, and we also wouldn't be able to use fd-passing to hook up stdin/stdout/stderr transparently. The next best thing would be something TCP-based like ssh, which means dealing with authentication and security; and the user of the IPC client would need to know whether it was expecting to be told to run a Linux process like `xterm` or a Windows process like `cmd`, in order to pass in the expected thing.

If Wine had more knowledge of the inner workings of the IPC server, then in principle I suppose we could give it a way to specify "the thing I want to run is a Windows executable":

```
steam-runtime-launch-client --bus-name=... -- xterm
steam-runtime-launch-client --bus-name=... --compatible-with=windows -- cmd.exe
```

and have it automatically turn that into `/path/to/wine64 cmd.exe`? But for that to work, Proton would have to pass options into `steam-runtime-launcher-service` to tell it how it can get its arguments to be run like a Windows executable instead of a Linux one, which undermines your second point.

smcv:
> Why does each compatibility tool have to decide to run via steam-runtime-launch-server on its own? Can we have some runner that would be used unconditionally, if present, and make the decision for itself if it sets up the server or just ends up doing execv?

I have wondered about this myself, but I think we should probably iterate a bit to make it as useful as possible, and then introduce a convenience entry point for "do the right thing" once we are more confident that we know what the right thing is. `s-r-launcher-service` is more mechanism than policy right now, with Steam and the compat tools like SLR and Proton providing the policy. When we are sure we've got the policy right, we can introduce a (versioned!) entry point in the `PATH` that makes it as simple as possible to implement that policy?

If we're running a wrapper unconditionally, then it's a single point of failure, and crash bugs in it become critical (preventing running *any* games, not just for developers but also for users), which does concern me.

Creating an "adverb" process like this is reasonably cheap in terms of startup times, but not free, which is why you see the generically-named `pressure-vessel-adverb` in the process hierarchy of SLR games: it combines all the things that need to wrap child processes (adjusting `LD_LIBRARY_PATH`, creating `ld.so.cache`, generating locales, holding a lock to stop the runtime being deleted prematurely, and so on) so that we don't have to pay the startup-time cost more times than we have to. So far, I've prototyped the SLR side of `s-r-launcher-service` integration as being part of the shell script that starts it up, but the shell script will get unmanageable if we're not careful, so I'm tempted to move more of the logic for whether to activate the debugging interface into the C code that starts up the container.

(I did consider combining the `launcher-server` with the `adverb`, but they have conflicting requirements in terms of Unix process management: the `launcher-server` wants to be an ordinary GLib program that runs a main loop and cooperates with other modules by only reaping known child processes in a targeted way, while the `adverb` is a subreaper and wants to reap arbitrary child processes with an indiscriminate `wait()`, and those programming models can't work together. `pidfd` promises to solve that incompatibility any year now...)

smcv:
> There should really be a better way of getting your own appid than having it hardcoded

I basically agree, but how would that work? It's easy for Steam to set environment variables that apply to the entire subprocess hierarchy, but hard to set them on a per-tool basis so that SLR_soldier and Proton each see their own correct app-ID. The least-bad thing I can think of is to insert a bunch of `env` commands into the command-line:

```
/path/to/reaper ... -- \
env STEAM_COMPAT_TOOL_ID=1391110 \
/path/to/SteamLinuxRuntime_soldier/_v2-entry-point --verb=waitforexitandrun -- \
env STEAM_COMPAT_TOOL_ID=1887720 \
"/path/to/Proton 7.0/proton" waitforexitandrun \
"/path/to/Floating Point.exe"
```

which seems quite nasty, and will be problematic for anything that parses the command-line and tries to pick it apart into compat tools (which nobody should be doing in production, but the `steam-runtime-launch-options` debugging tool has to do this, because the way Steam implements launch options doesn't really leave it any alternative).

Using `env` like this would result in "leaking" the last compat tool's ID into the final game process: we can't clear it out with `env -u STEAM_COMPAT_TOOL_ID`, because Proton's argument is a Windows executable rather than a Linux command, and other compat tools could in principle have arguments that are neither (like a hypothetical ScummVM compat tool might take a ScummVM game-ID or directory name rather than the path to an executable). That "leak" is maybe not so bad, because it isn't awful for the game to know what compat tool it's running under.

Also, if one of the executable names might have an equals sign in it (for example if a user's Steam library is `/inadvisable=path`), the documented way to avoid `env` misbehaving is this rather obscure rune:

```
/path/to/reaper ... -- \
env STEAM_COMPAT_TOOL_ID=1391110 \
sh -c 'exec "$@"' sh \
/inadvisable=path/steamapps/common/SteamLinuxRuntime_soldier/_v2-entry-point --verb=waitforexitandrun -- \
env STEAM_COMPAT_TOOL_ID=1887720 \
sh -c 'exec "$@"' sh \
"/inadvisable=path/steamapps/common/Proton 7.0/proton" waitforexitandrun \
"/inadvisable=path/Floating Point.exe"
```

(Arguably it should be `exec -- "$@"` to avoid brokenness if the executable name somehow starts with `-`, but that only works in bash (which would make it slower and less robust), and not in a more compact POSIX `/bin/sh` like the one in Debian/Ubuntu. Isn't portability great?)

proton
smcv: Sorry, I'm still unclear about which of the scripts in `tmpdir` are intended to be run from inside the container, and which are intended to be run from outside?

If they're intended to be run from inside the container and that is part of their "API", then I should probably add a version that is intended to be run from outside (and therefore uses `s-r-launch-client` for IPC) instead. If they're already intended to be run from outside, then I can change them like I did for `run` in the initial version of this MR.

The debug scripts are:

* `winedbg`: basically `winedbg "$@"`
* `winedbg_run`: same as `winedbg`, but if no arguments are given, the default is to run the actual game
* `gdb_attach`: tries to find an executable of a given name and run `winedbg --gdb` against it
* `gdb_run`: same as `winedbg_run` but with `--gdb`
* `run`: run arguments or the actual game, wrapped in `steam.exe`

proton
smcv: > If we had a mode where steam-runtime-launcher-service can be told to keep running after the command it wraps has exited, which is something I plan to implement at some point, then that would make this script more useful.

I implemented this, it'll probably be in the next container beta (run with `SRT_LAUNCHER_SERVICE_STOP_ON_EXIT=0`).

proton
smcv: I've added a TODO comment to say so.

proton
smcv: Since I'm still unsure about this, I've moved this part into a second commit, so that the bit I'm more confident about (the first commit) can be considered separately.

In the second commit, I've changed it to use IPC for `winedbg` and `winedbg_run` but not `run`. Other suggestions welcome...

ivyl:
> You're right that this is a bit odd, but the s-r-launcher-service is a Linux executable, using Linux mechanisms for IPC (D-Bus over AF_UNIX, with Unix-specific fd-passing to tunnel the stdin/stdout/stderr file descriptors from outside the container), so you can't run it as a Windows executable. (Or can you? I don't know what happens if a Windows executable being run via Wine tries to run a Linux executable by its path or by PATH search.)

I think what we are doing here is the most sensible option for developers, even though it creates a bit of dissonance and requires knowledge about Proton/Wine.

> I have wondered about this myself, but I think we should probably iterate a bit to make it as useful as possible, and then introduce a convenience entry point for "do the right thing" once we are more confident that we know what the right thing is. 

It would be still better to have the entry point already so we can do something like

`steam-runtime-launcher-service-smart --tool-name=proton -- ...` and it would check `STEAM_COMPAT_LAUNCHER_SERVICE`, handle setting `bus-name` based on app id etc.

If this is too scary to call it unconditionally then we can hide it behind an extra check of `STEAM_COMPAT_LAUNCHER_SERVICE` inside `proton` and remove that check later, once we have enough confidence.

> I basically agree, but how would that work? It's easy for Steam to set environment variables that apply to the entire subprocess hierarchy, but hard to set them on a per-tool basis so that SLR_soldier and Proton each see their own correct app-ID.

Then maybe we should not assume that the app knows the ID and instead we can rely solely on the human-friendly names like `proton` or `container-runtime`? If Steam / other tools needs to know them we can just add those to `toolmanifest.vdf`.

ivyl:
```diff
diff --git a/proton b/proton
index 4bf158b5..06629194 100755
--- a/proton
+++ b/proton
@@ -1450,9 +1484,14 @@ class Session:
         return subprocess.call(args, env=local_env, stderr=self.log_file, stdout=self.log_file)
 
     def run(self):
+        if shutil.which('steam-runtime-launcher-service') is not None:
+            adverb = ['steam-runtime-launcher-service', '--tool-name', 'proton', '--' ]
+        else:
+            adverb = []
+
         if "PROTON_DUMP_DEBUG_COMMANDS" in self.env and nonzero(self.env["PROTON_DUMP_DEBUG_COMMANDS"]):
             try:
@@ -1467,9 +1506,11 @@ class Session:
 
         # CoD: Black Ops 3 workaround
         if os.environ.get("SteamGameId", 0) == "311210":
-            rc = self.run_proc([g_proton.wine_bin, "c:\\Program Files (x86)\\Steam\\steam.exe"] + sys.argv[2:] + self.cmdlineappend)
+            argv = [g_proton.wine_bin, "c:\\Program Files (x86)\\Steam\\steam.exe"]
         else:
-            rc = self.run_proc([g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"] + sys.argv[2:] + self.cmdlineappend)
+            argv = [g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"]
+
+        rc = self.run_proc(adverb + argv + sys.argv[2:] + self.cmdlineappend)
 
         if remote_debug_proc:
             remote_debug_proc.kill()
diff --git a/toolmanifest_runtime.vdf b/toolmanifest_runtime.vdf
index 0b8490e6..8c32fcb1 100644
--- a/toolmanifest_runtime.vdf
+++ b/toolmanifest_runtime.vdf
@@ -1,6 +1,7 @@
 "manifest"
 {
   "version" "2"
+  "name" "proton"
   "commandline" "/proton %verb%"
   "require_tool_appid" "1391110"
   "use_sessions" "1"

```

This is more along the lines what I would like to see. We don't need to know our app id and instead Steam / other tooling can discover our developer-friendly name via toolmanifest.

Also here `steam-runtime-launcher-service` does the heavy lifting in terms of checking `STEAM_COMPAT_LAUNCHER_SERVICE`, deciding on the dbus address and printing messages directed to the developer.  This way we don't have to duplicate that logic between SRT and Proton.

If it's too scary we can add `and self.env.get("STEAM_COMPAT_LAUNCHER_SERVICE", "") == "proton"` for the time being.

smcv:
> We don't need to know our app id and instead Steam / other tooling can discover our developer-friendly name via toolmanifest.

I'll have to defer to @TTimo on whether that's feasible within Steam, I don't have access to Steam itself.

smcv:
> Also here steam-runtime-launcher-service does the heavy lifting in terms of checking STEAM_COMPAT_LAUNCHER_SERVICE, deciding on the dbus address and printing messages directed to the developer. This way we don't have to duplicate that logic between SRT and Proton.

If SRT and Proton are invoking something *unconditionally*, I don't really want the thing they invoke to be steam-runtime-launcher-service, because it has non-trivial code and dependencies, which could easily make all SLR and Proton games slow or non-functional; and because that isn't a versioned name, which will give us problems if we want an incompatible behaviour or interface change.

Having a more trivial entry point with a versioned name like `steam-runtime-launcher-interface-0`, which execs either the full-fat `steam-runtime-launcher-service` or the wrapped command itself, could be OK. I'd be tempted to go as far as not giving it getopt-style command-line processing at all, so `argv[1..]` are always just the wrapped command, and any behaviour-changing parameters have to come in via the environment or something (and if we later decide that any of this was a mistake, `steam-runtime-launcher-interface-1` can fix it without worrying about having to be 100% backwards-compatible).

I suspect we'll go through a few interface versions initially, because I'll have to guess how it should behave with only a limited ability to prove that it works for real-world use-cases, but perhaps I'm being pessimistic.

ivyl:
>>  We don't need to know our app id and instead Steam / other tooling can discover our developer-friendly name via toolmanifest.

> I'll have to defer to @TTimo on whether that's feasible within Steam, I don't have access to Steam itself.

Fingers crossed. That would make our life much easier with the many, many apps we have for Proton and the occasional manual build :-)

> Having a more trivial entry point with a versioned name like steam-runtime-launcher-interface-0, which execs either the full-fat steam-runtime-launcher-service or the wrapped command itself, could be OK. I'd be tempted to go as far as not giving it getopt-style command-line processing at all, so argv[1..] are always just the wrapped command, and any behaviour-changing parameters have to come in via the environment or something (and if we later decide that any of this was a mistake, steam-runtime-launcher-interface-1 can fix it without worrying about having to be 100% backwards-compatible).

Sounds like a not too bad of an option. I imagine that we would need to pass the tool name somehow, either as a environment variable, like you have suggested, or the first parameter.

```diff
diff --git a/proton b/proton
index 4bf158b5..06629194 100755
--- a/proton
+++ b/proton
@@ -1450,9 +1484,14 @@ class Session:
         return subprocess.call(args, env=local_env, stderr=self.log_file, stdout=self.log_file)
 
     def run(self):
+        if shutil.which('steam-runtime-launcher-interface-0') is not None:
+            adverb = ['steam-runtime-launcher-service', 'proton']
+        else:
+            adverb = []
+
```

> If SRT and Proton are invoking something unconditionally, I don't really want the thing they invoke to be steam-runtime-launcher-service, because it has non-trivial code and dependencies, which could easily make all SLR and Proton games slow or non-functional; and because that isn't a versioned name, which will give us problems if we want an incompatible behavior or interface change.

I'm not completely opposed to keeping `if self.env.get("STEAM_COMPAT_LAUNCHER_SERVICE", "") == "proton"` in addition to / instead of `shutil.which()`, but past that it would be nice if the service/interface would handle the rest.

Out of curiosity, why does the bus name include the SteamAppId of the game?

smcv:
> ```
> +        if shutil.which('steam-runtime-launcher-interface-0') is not None:
> +            adverb = ['steam-runtime-launcher-service', 'proton']
> ```

I meant more like `adverb = ['steam-runtime-launcher-interface-0', 'proton']`.

smcv:
> Out of curiosity, why does the bus name include the SteamAppId of the game?

It's not clear to me whether Steam can run more than one game at a time; and if it can't, it's not clear to me whether that's considered to be a bug that might be fixed in a future Steam release.

The bus name being non-unique is not *necessarily* a fatal problem with the latest steam-runtime-tools release, which has a `--no-stop-on-name-loss` option (currently not the default, but I might make it the default for `steam-runtime-launcher-interface-0`), because the well-known name will go to the most-recently-run launcher-service, and you can connect to the others via their unique name like `:1.42`; but that's not very discoverable.

TTimo:
> > > We don't need to know our app id and instead Steam / other tooling can discover our developer-friendly name via toolmanifest.
> 
> > I'll have to defer to @TTimo on whether that's feasible within Steam, I don't have access to Steam itself.
> 
> Fingers crossed. That would make our life much easier with the many, many apps we have for Proton and the occasional manual build :-)
> 

Having a human friendly / common name carried in the manifest should be easy.

I've read through the rest of the discussion and I generally don't have concerns. Please pick a design (using adverbs `steam-runtime-launcher-interface-0`, with or without env vars) and provide some beta builds of SLR (or patches where adequate) so I can iterate on the Steam side?

ivyl:
> Having a human friendly / common name carried in the manifest should be easy.

Great! So we can just add the "name" field? The toolmanifest has "version" field and I wonder if we need to touch that or not.

> I meant more like `adverb = ['steam-runtime-launcher-interface-0', 'proton'].`

I'll cook up a version using this.

TTimo:
@smcv I think you'll need to do similar changes in SLR?

smcv:
> @smcv I think you'll need to do similar changes in SLR?

Yes, and I'll also need to write the versioned entry-point. I'll hopefully start on that tomorrow.

proton
smcv: This is no longer needed in today's revision.

proton
smcv: In today's revision I've stopped editing those scripts. We can't make them directly runnable from the host unless/until Proton knows how to find a path to (a recent version of) `steam-runtime-launch-client` *that is valid for the host* (the one that Proton finds in `PATH` will usually only exist at that path inside the container).

If you want these scripts to be directly runnable from the host, we'll probably have to wait for a sufficiently new `steam-runtime-launch-client` to be included in the Steam client itself, and then have the scripts invoke it as `~/.steam/root/ubuntu12_32/steam-runtime/amd64/bin/steam-runtime-launch-client`.

Alternatively, users can run them like this:

```
/path/to/steam-runtime-launch-client --list
/path/to/steam-runtime-launch-client --bus-name=... -- /tmp/proton_$(id -nu)/winedbg
```

proton
smcv: Now `steam-runtime-launch-interface-0` shows this information itself (it'll end up in whatever log is receiving its stderr at the time).

proton
smcv: I'm no longer altering these scripts.

smcv:
> I meant more like `adverb = ['steam-runtime-launcher-interface-0', 'proton']`.

Implemented, and seems to work. You'll need a bit of juggling components to make it work before this stuff appears in public releases (see edited MR description), but surprisingly little of that is necessary.

smcv:
>>>
> ```diff
> --- a/toolmanifest_runtime.vdf
> +++ b/toolmanifest_runtime.vdf
> @@ -1,6 +1,7 @@
>  "manifest"
>  {
>    "version" "2"
> +  "name" "proton"
>    "commandline" "/proton %verb%"
>    "require_tool_appid" "1391110"
>    "use_sessions" "1"
> ```

I think we should have a more descriptive field name here than just `name`.

The semantics that we want for it are:

* it's a short, programmer-understandable token like `proton`, not a user-visible "pretty name" like "Proton - Experimental"
* it's *not* unique across all compat tools, but only describes the role it plays within a stack: Proton 6, Proton 7, Proton Experimental, unofficial Proton builds like GE, etc. are all `proton`, and soldier, sniper etc. are all `container-runtime`.
* whenever we build a stack of compat tools, it's unique within that stack, so that we can address any single layer of the stack using one of these tokens

Having Proton-in-Proton would make no sense, so all Protons can safely have the same token here, and nesting containers doesn't work, so all container runtimes can safely have the same token here.

I'm not sure what expresses that best. Perhaps something involving `role` or `layer`?

TTimo:
how about `compatmanager_layer_name` ?

proton
ivyl: I'll have a look into this. Sounds sensible, but I wonder if there's a reason to have it like that. Only thing that comes to mind is to have proton easily identifiable in the process tree.

ivyl:
Thanks! I've added a commit that sets the compatmanager_layer_name. The PR is okay to merge, I'm happy about how it has shaped up :-)

I think it should be safe to have it in bleeding-edge already. Let me know if you want it merged now or wait a bit.

As of the debug scripts - I still see value in having them dumping the whole custom env, e.g. to use with standalone wine. It still has some use for outside-of-SRT debugging that Proton developers do. I agree it's not perfect and it's a more of advanced topic. There's a long overdue cleanup done too. I'll handle that separately.

As of the second `toolmanifest_noruntime.vdf` I don't see a reason for keeping it. I think it's an artifact from times when SRT was still novel. That will also be cleaned up as a follow up on my end

TTimo:
Not directly related to this, but note that we'd like to do away with the toolmanifest vdf that is embedded in the compat tool files eventually. It's a remnant from early implementation and causes some minor problems, we'd like to consolidate that to the appinfo.

smcv:
> Not directly related to this, but note that we'd like to do away with the toolmanifest vdf that is embedded in the compat tool files eventually. It's a remnant from early implementation and causes some minor problems, we'd like to consolidate that to the appinfo.

For SLR (and probably Proton too) we'll need to be quite careful that the tool manifest stays in sync with the actual capabilities and behaviour of the compat tool, including across upgrades, and in situations where different branches have different (incompatible) behaviour. I know we've had some trouble with this in the past, where Steam updated one of the SLR depots to a version with incompatible behaviour, but didn't reload `toolmanifest.vdf` until you exit/restart Steam.

Or would the proposed change make Steam *more* able to synchronize its opinion on each compat tool with the reality, by having the tool manifest live somewhere that Steam can reload more regularly?

ivyl (approved):

ivyl:
It's now merged into our staging experimental_7.0 branch and is already live on experimental's `bleeding-edge` steam beta.

https://github.com/ValveSoftware/Proton/commits/experimental-bleeding-edge-7.0-20939-20220727-p35943e-wfd43ae-df5fe25-vd00d03

It will make it into the next release of experimental and 7.0 dash release. Thanks!

smcv:
@ivyl FYI, the necessary pressure-vessel version has also been released into the `client_beta` branch of "Steam Linux Runtime - soldier" (and also sniper, if anyone is already working on a Proton version that uses that).

The requirement is that `VERSIONS.txt` says `pressure-vessel 0.20220726.0` or later.

## Steps to test

Run Steam from a terminal with no special options or environment variables.

Opt in to Proton Experimental `[bleeding-edge]` branch and Steam Linux Runtime - soldier `[client_beta]` branch. (This step will stop being necessary as these changes go through the release pipeline.)

Verify integrity (just to make sure no temporary hacks are still in place).

Have any Proton game set to use Proton Experimental. I used The Expendabros (312990) which is small and free-to-play, so anyone can follow these steps.

Put `STEAM_COMPAT_LAUNCHER_SERVICE=proton` in the game's Launch Options (this step will be replaced by a Steam feature after @TTimo has integrated it).

Launch the game.

This text appears on Steam's stderr:

```
…

Starting program with command-launcher service.

To inject commands into the container, use a command like:

/home/me/SteamLibrary/steamapps/common/SteamLinuxRuntime_soldier/pressure-vessel/bin/steam-runtime-launch-client \
        --bus-name=:1.430 \
        --directory='' \
        -- \
        bash

esync: up and running.
wine: RLIMIT_NICE is <= 20, unable to use setpriority safely
…
```

(Obviously the Steam Library path /home/me/SteamLibrary and the unique name :1.430 will vary.)

Run the indicated command in another terminal. You get an interactive shell inside the container. Its current working directory is the game's cwd (in my case `.../Broforce The Expendables Missions`). `/etc/debian_chroot` and `/usr/lib/os-release` indicate that it is soldier.

`env | LC_ALL=C sort | grep ^WINE` indicates that the Proton environment variables are set, notably `$WINEPREFIX`. Also, `Proton - Experimental/files/bin` is in the `PATH`, so I can run commands like `wine64 winedbg notepad.exe` with no need for additional runes.

If the command printed on stderr is not convenient to retrieve (perhaps it's going to a log or the Journal or /dev/null), another way to get into the container is to run:

```
.../SteamLinuxRuntime_soldier/pressure-vessel/bin/steam-runtime-launch-client --list
```

and it will show you a list of `--bus-name` options you can substitute into a similar command.

For completeness: if you use `STEAM_COMPAT_LAUNCHER_SERVICE=container-runtime` instead of `proton`, instead you'll get a shell in the container environment, "outside" Proton. You can use this to debug the `./proton` script if you want.

## Errata

At the moment, if you want to be in the game's working directory, you have to use `--directory=""` (i.e. an empty string). The default is basically `--directory="$(pwd)"`, using the current working directory of the shell where you run `s-r-launch-client`. This is inherited from `flatpak-spawn` (from which `s-r-launch-client` was derived) and is probably the wrong default, so I might change it (`steamrt/tasks#145`); but if you explicitly use `--directory=""` then that should always be safe.

It's possible to use `~/.steam/root/ubuntu12_32/steam-runtime/amd64/usr/bin/steam-runtime-launch-client` as an alternative to the one bundled in pressure-vessel, but it's currently only present in the Steam beta, which currently only has an older version of `s-r-launch-client`. This older version is lacking many quality-of-life features like pseudo-terminal support (required for shell job control) and `--list`.

Issue #5897, opened 2022-06-12
POSTAL: Brain Damaged (1359980)


Bitwolfies:
# Compatibility Report
- Name of the game with compatibility issues:POSTAL: Brain Damaged 
- Steam AppID of the game:1359980

## System Information
- GPU: GTX 3080 
- Driver/LLVM version: Nvidia 515 DKMS
- Kernel version: 5.18
- Proton version: 7.0

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

## Symptoms 
Videos by default play corrupted, this can be solved by adding WINE_DO_NOT_CREATE_DXGI_DEVICE_MANAGER=1 %command% to launch, needs to be added to the hack config in proton, ironically like Postal 4.

Beyond that, the game works perfectly.

Logge1002:
First stage all the windows have broken reflections please fix that!

gander-code:
> First stage all the windows have broken reflections please fix that!

I'm having this issue as well, not sure what caused it but it used to work.

Gibitwit:
Reflection cube maps do not render correctly.  They appear to be placed on surfaces as a regular flat texture.  Looking on the Steam forum for this game every Linux player seems to have this problem.

PR #5898, opened 2022-06-12
proton: Add Postal: Brain damged to MFDXGI hack.

Bitwolfies:
CW-Bug-Id: #19126

Related issue: https://github.com/ValveSoftware/Proton/issues/5897


kisak-valve:
Minor nitpick, typo: "damged" in the first commit, and "Added demo" isn't very clear on its own. Repeating the first commit title with it mentioning the demo or combining the two commits would be better.

Issue #5905, opened 2022-06-14
Terra Nil Demo (1642370)


Newbytee:
# Compatibility Report
- Name of the game with compatibility issues: Terra Nil Demo
- Steam AppID of the game: 1642370

## System Information
- GPU: Intel HD 5000 (though I can see people having the same issue with NVIDIA and AMD GPUs on ProtonDB)
- Driver/LLVM version: Mesa 21.1.1
- Kernel version: 5.17.13
- Link to full system information report as [Gist](https://gist.github.com/):
- Proton version: Latest Experimental at the time of writing 

## I confirm:
- [X] that I haven't found an existing compatibility report for this game.
- [X] that I have checked whether there are updates for my system available.

Log: [steam-1642370.log](https://github.com/ValveSoftware/Proton/files/8901033/steam-1642370.log) (both the corrupted video issue and the freeze issue were reproduced)

## Symptoms 

Certain videos in the game are garbled, e.g. the main menu background one:

![image](https://user-images.githubusercontent.com/4428753/173612236-749eb75c-0b07-4bf3-a1b4-084dd9675924.png)

The game also hangs when I tab out of it.

## Reproduction

Video issue:
1. Start game
2. Observe that the main menu background is corrupted
3. Launch into the gameplay
4. Observe that tooltip videos are garbled in the same manner

Freeze:
1. Start game
2. Tab out (using GNOME Wayland in my case)
3. Observe that game freezes

kisak-valve:
Hello @Newbytee, the Intel HD 5000 chipset is part of the Haswell generation. The video drivers for that generation does not provide a spec-conformant implementation of Vulkan. The freeze you're seeing is most likely a video driver issue which might be worth mentioning to the mesa/ANV devs.

That said, can you test how the game behaves with the launch options set to `WINE_DO_NOT_CREATE_DXGI_DEVICE_MANAGER=1 %command%`?

Blisto91:
I was not able to reproduce the alt tab freeze from a quick couple of tries on my amd R9 380 mesa 22.1.1.
Tho i can verify that `WINE_DO_NOT_CREATE_DXGI_DEVICE_MANAGER=1 %command%` fixes the main menu video.

If you are able to then try to update your graphics drivers.

Newbytee:
[...]

PR #5907, opened 2022-06-14
proton: Add Terra Nil Demo to MFDXGI hack.

Newbytee:
CW-Bug-Id: #19126

Works around main menu background video being corrupted.

See https://github.com/ValveSoftware/Proton/issues/5905


kisak-valve:
Nitpick: This appid is for Terra Nil Demo, not the full game release.

Newbytee:
Thanks for the commend, fixed.

ivyl:
Hi. Thanks for the patch. I've pushed it to our staging tree and is should be included in the next experimental release.

PR #5922, opened 2022-06-19
Respect VR_OVERRIDE envvar

Supreeeme:
Useful for alternative OpenVR runtimes, such as OpenComposite.


aeikum:
Thanks, looks good. I've merged this in and it will be in an upcoming Experimental release as well as the next 7.0 release.

PR #5964, opened 2022-07-03
Proton build with GitHub actions [proton_7.0 branch]

dylanmtaylor:
This pull request adds a GitHub action to build the project on each commit leveraging GitHub's CI environment. The resultant redist package is then uploaded as an artifact which can be downloaded from the actions tab. This PR would have to be merged into each branch for it to run there. If accepted, this should be cherry-picked to every branch.

Here's what this action looks like in practice, being run on the proton 7.0 branch: https://github.com/dylanmtaylor/Proton/actions/runs/2609849229

GitHub Actions CI minutes are free to public and open source repositories like Proton. :)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "wine"]
 	path = wine
-	url = ../wine
+	url = https://github.com/ValveSoftware/wine
 [submodule "dxvk"]
 	path = dxvk
 	url = https://github.com/doitsujin/dxvk.git
--- a/.github/workflows/build_proton.yml
+++ b/.github/workflows/build_proton.yml
@@ -1,0 +1,52 @@
+name: Build Proton
+
+# Allows manual runs, and runs on each commit or release
+on: [workflow_dispatch, push, release]
+
+jobs:
+  make:
+    runs-on: ubuntu-latest
+    name: Make
+    strategy:
+      matrix:
+        build_type: ['redistributable', 'deployment']
+        build_strip: ['stripped', 'unstripped']
+    steps:
+      - name: Checkout the repository
+        id: checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          lfs: true
+
+      - name: Install build dependencies
+        id: dependencies
+        run: sudo apt-get -y install fontforge
+
+      - name: Build proton and generate a ${{ matrix.build_type }} package
+        id: make
+        run: |
+          if [[ "${{ matrix.build_strip }}" == "unstripped" ]]; then
+            export UNSTRIPPED_BUILD=1
+          fi
+          
+          if [[ "${{ matrix.build_type }}" == "redistributable" ]]; then
+            make redist
+          elif [[ "${{ matrix.build_type }}" == "deployment" ]]; then
+            make deploy
+          else
+            exit 1
+          fi
+
+      - name: Recursively list directory contents for debugging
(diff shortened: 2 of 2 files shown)


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "wine"]
 	path = wine
-	url = ../wine
+	url = https://github.com/ValveSoftware/wine
```
dylanmtaylor, line 3: The reason this needs to be an absolute path is for forks so that the fork doesn't also need a fork of wine.

dylanmtaylor: I suppose the origin path could also be set in CI so this isn't needed.

ivyl:
Hi. I've briefly looked into GitHub's CI actions about a year ago and decided that it's way too slow.

We have automation already in place using GitLab we use for the development. It has a dedicated, trusted runner for building. Builds take ~3min per artifact created which is much better than the ~2h with GitHub actions.

We've been investing time into making build as simple as possible. Since it's `make build_name=xyz deploy` there's nothing worth sharing in this regard. Most of the logic is around automated Steam uploads (bleeding-edge, -rcs, builds for QA testing before public release, etc.) and tagging (the tags are pushed publicly, see bleeding-edge).

dylanmtaylor:
> Hi. I've briefly looked into GitHub's CI actions about a year ago and decided that it's way too slow.
> 
> We have automation already in place using GitLab we use for the development. It has a dedicated, trusted runner for building. Builds take ~3min per artifact created which is much better than the ~2h with GitHub actions.
> 
> We've been investing time into making build as simple as possible. Since it's `make build_name=xyz deploy` there's nothing worth sharing in this regard. Most of the logic is around automated Steam uploads (bleeding-edge, -rcs, builds for QA testing before public release, etc.) and tagging (the tags are pushed publicly, see bleeding-edge).

Only 3 minutes per artifact? That's incredible. Even though I spent a bit of time on this, I understand if you want to close it because this doesn't provide you with a great amount of benefit over your current system. I'm guessing that GitLab works with cached build artifacts for the upstream projects to save time while this builds the whole thing. Or if you want you can merge this in and have GitHub Actions as a backup. The artifacts being on the individual commits and tags here is kind of nice.

ivyl:
> Only 3 minutes per artifact? That's incredible.

Dedicated server with 18 cores / 36 threads and tons of RAM is the biggest contributor. Building on a ramdisk and having the repos pre-cloned and just fetching the changes helps a lot.

> I'm guessing that GitLab works with cached build artifacts for the upstream projects to save time while this builds the whole thing.

The machine being a dedicated one helps the most - not discarding the podman, cargo and ccache caches. Also not having to install the few host-side dependencies every time (FWIW we are working on having everything in-container). As a developer you get all of those benefits on your desktop out of the box as this is what the `Makefile` defaults to :-)

It's just that it's all squandered when using public CI infrastructure where nothing is really persistent. And you don't get even close to a 18C/36T machine.

> Even though I spent a bit of time on this, I understand if you want to close it because this doesn't provide you with a great amount of benefit over your current system.

> Or if you want you can merge this in and have GitHub Actions as a backup. The artifacts being on the individual commits and tags here is kind of nice.

Thanks for looking into this. I don't think it's worth duplicating the CI setup and having another source of Proton binaries. It's also a maintenance burden. I'll looks into including redist builds created by the CI when making releases though.

PR #6120, opened 2022-08-26
Enable dxvknvapi on a bunch of (DLSS) titles I've verified as functioning and stable.

adamnv:
Note: in some cases I've included the app ID for a game demo but not the full game.  This is only because in those cases I've not tested the full game, and wanted to strictly only include titles I've directly verified as stable'n'good.

However, it's probably very fair to say that if a game demo had good DLSS/nvapi stability then the full game will also have good DLSS/nvapi stability, so if you'd like me to add the app IDs for full titles sight-unseen then I'm happy to do so!  **(edit: done that.)**

### DIFF
--- a/proton
+++ b/proton
@@ -1058,6 +1058,7 @@ def default_compat_config():
                 "870780", #control ultimate edition
                 "1091500", #cyberpunk 2077
                 "1190460", #death stranding
+                "1850570", #Death Stranding Director's Cut
                 "1252330", #deathloop
                 "548430", #deep rock galactic
                 "428660", #deliver us the moon
@@ -1067,6 +1068,10 @@ def default_compat_config():
                 "1952070", #engine evolution 2022 demo
                 "1128920", #everspace 2
                 "1312800", #everspace 2 demo
+                "1330470", #F.I.S.T.: Forged In Shadow Torch
+                "1332390", #F.I.S.T.: Forged In Shadow Torch Demo
+                "1641960", #Forever Skies
+                "2141060", #Forever Skies Demo
                 "1080110", #f1 2020
                 "1098130", #get stuffed
                 "1139900", #ghostrunner
@@ -1079,6 +1084,8 @@ def default_compat_config():
                 "1987940", #island of the ancients demo
                 "1371480", #iron conflict
                 "1544360", #lego builder's journey
+                "1363080", #Manor Lords
+                "2122820", #Manor Lords Demo
                 "1817070", #marvel's spider-man remastered
                 "784080", #mechwarrior 5: mercenaries
                 "1170950", #mortal online 2
--- a/proton
+++ b/proton
@@ -1042,6 +1042,71 @@ def default_compat_config():
                 ]:
             ret.add("noopwr")
 
+        if appid in [
+                # enable dxvknvapi for titles verified to benefit (e.g. working DLSS)
+                "673130", #amid evil
+                "1291680", #apocalypse: 2.0 edition
+                "979690", #the ascent
+                "805550", #assetto corsa competizione
+                "924970", #back 4 blood
+                "1178830", #bright memory infinite
+                "1409670", #bright memory infinite benchmark
+                "1153640", #chorus
+                "1791040", #chorus demo
+                "1577240", #cions of vega
+                "1632760", #cions of vega demo
+                "870780", #control ultimate edition
+                "1091500", #cyberpunk 2077
+                "1190460", #death stranding
+                "1252330", #deathloop
+                "548430", #deep rock galactic
+                "428660", #deliver us the moon
+                "534380", #dying light 2
+                "269190", #edge of eternity
+                "1871990", #engine evolution 2022
+                "1952070", #engine evolution 2022 demo
+                "1128920", #everspace 2
+                "1312800", #everspace 2 demo
+                "1080110", #f1 2020
+                "1098130", #get stuffed
+                "1139900", #ghostrunner
+                "1249200", #ghostrunner demo
+                "1593500", #god of war
+                "414340", #hellblade: senua's sacrifice
+                "1659040", #hitman 3
+                "1847520", #hitman 3 free starter pack
+                "1650150", #island of the ancients
+                "1987940", #island of the ancients demo
+                "1371480", #iron conflict
+                "1544360", #lego builder's journey
+                "1817070", #marvel's spider-man remastered
+                "784080", #mechwarrior 5: mercenaries
+                "1170950", #mortal online 2
+                "261550", #mount & blade II: bannerlord
+                "275850", #no man's sky
+                "1140100", #the persistence
+                "1322170", #pluviophile
+                "1186640", #pumpkin jack
+                "391220", #rise of the tomb raider
+                "750920", #shadow of the tomb raider
+                "1602080", #soulstice
+                "2015300", #soulstice demo
+                "1296010", #stay in the light
+                "813630", #supraland
+                "487390", #system shock demo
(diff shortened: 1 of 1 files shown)


ivyl:
It would be good to have also the full versions of the game tested, but that's something we can add based on the demos.

Just to understand the extent of testing you have done, have you tried any of the games with older, non-RTX GPUs? 

It should not negatively affect games running with other vendors' GPUs but it would be nice to make sure that's the case.

The change would go into experimental which already requires VK_KHR_dynamic_rendering, so that limits the scope of the backwards compatibility we would need to check for :-)

adamnv:
Thanks for the reponse!  Comments inline:

> It would be good to have also the full versions of the game tested, but that's something we can add based on the demos.

I hope to (re)test just about everything in the fullness of time (and steam sales :laughing:) but I'm happy to add the non-demos on the basis of the success with the demos; are you saying I should go ahead with that in this PR?
 
> Just to understand the extent of testing you have done, have you tried any of the games with older, non-RTX GPUs?

On the one hand - no, I haven't!  On the other hand, I'm familiar enough with the failure modes of the only dxvknvapi-enabled tech which overlaps with RTX cards - DLSS - to be adequately confident that DLSS itself fails gracefully.
 
> It should not negatively affect games running with other vendors' GPUs but it would be nice to make sure that's the case.

I agree; FWIW I believe dxvknvapi is designed by default to do nothing outside of the NVIDIA driver stack unless a user is explicitly spoofing a customVendorId, but I'm checking that assumption again with @jp7677 :grin:  edit: confirmed by Jens - dxvknvapi does nothing outside of the NVIDIA driver stack by default, regardless of PROTON_ENABLE_NVAPI.

> The change would go into experimental which already requires VK_KHR_dynamic_rendering, so that limits the scope of the backwards compatibility we would need to check for :-)

Would you like me to retarget the PR against another branch or is this one okay?

jp7677:
> I agree; FWIW I believe dxvknvapi is designed by default to do nothing outside of the NVIDIA driver stack unless a user is explicitly spoofing a customVendorId, but I'm checking that assumption again with @jp7677 grin edit: confirmed by Jens - dxvknvapi does nothing outside of the NVIDIA driver stack by default, regardless of PROTON_ENABLE_NVAPI.

Yes correct. When no found device runs on the NVIDIA proprietary driver, `NvAPI_Initialize` fails with `NVAPI_NVIDIA_DEVICE_NOT_FOUND`. This should match behavior on Windows. 

(You can override the default behavior with DXVK_NVAPI_ALLOW_OTHER_DRIVERS=1 for e.g. using Reflex or NVAPI D3D11 extensions with AMD cards, see https://github.com/jp7677/dxvk-nvapi#tweaks-debugging-and-troubleshooting. Setting PROTON_ENABLE_NVAPI does not touch this switch.)

> On the one hand - no, I haven't! On the other hand, I'm familiar enough with the failure modes of the only dxvknvapi-enabled tech which overlaps with RTX cards - DLSS - to be adequately confident that DLSS itself fails gracefully.

From my observations, before starting DLSS, applications query the architecture (Turing, Ampere etc) of the active card using NVAPI and do not continue firing up DLSS if the returned architecture is less than Turing, So this should indeed be safe on non-RTX cards.

jp7677:
It would be cool to also include APP-ID `805550` / `Assetto Corsa Competizione` (https://github.com/ValveSoftware/Proton/issues/1420) in this list. This game is the reason i started DXVK-NVAPI and serves as my standard test bed :). This game is based on UE4 and runs perfectly with DLSS and also uses D3D11 extensions (SetDepthBoundsTest/UAVOverlap).

adamnv:
> It would be cool to also include APP-ID `805550` / `Assetto Corsa Competizione` (#1420) in this list. This game is the reason i started DXVK-NVAPI and serves as my standard test bed :). This game is based on UE4 and runs perfectly with DLSS and also uses D3D11 extensions (SetDepthBoundsTest/UAVOverlap).

added - thanks!

ivyl:
> > It would be good to have also the full versions of the game tested, but that's something we can add based on the demos.
> 
> I hope to (re)test just about everything in the fullness of time (and steam sales laughing) but I'm happy to add the non-demos on the basis of the success with the demos; are you saying I should go ahead with that in this PR?

Yes. Adding all of them sounds good. Then we can knock off any games that prove to be problematic as we sift through them.

> > Just to understand the extent of testing you have done, have you tried any of the games with older, non-RTX GPUs?
> 
> On the one hand - no, I haven't! On the other hand, I'm familiar enough with the failure modes of the only dxvknvapi-enabled tech which overlaps with RTX cards - DLSS - to be adequately confident that DLSS itself fails gracefully.

I don't have a reason to distrust you here but I also don't fully trust any technology :-P

I think that landing it in bleeding-edge (the staging tree) would be fine but we will do some sanity checks on this before it makes its way to experimental or stable.

> > It should not negatively affect games running with other vendors' GPUs but it would be nice to make sure that's the case.
> 
> I agree; FWIW I believe dxvknvapi is designed by default to do nothing outside of the NVIDIA driver stack unless a user is explicitly spoofing a customVendorId, but I'm checking that assumption again with @jp7677 grin edit: confirmed by Jens - dxvknvapi does nothing outside of the NVIDIA driver stack by default, regardless of PROTON_ENABLE_NVAPI.

The comment I made above applies here as well.

> Would you like me to retarget the PR against another branch or is this one okay?

This one is okay. I apply the PRs manually anyway to the staging tree.

adamnv:
Cool - I'll get on it today.

adamnv:
> Yes. Adding all of them sounds good.

Done.  Cheers.

adamnv:
I've pushed a commit adding 7 more tested titles.

Going forward, should I continue to extend this PR/branch on an ongoing basis, or issue a new PR for every update?   Or something else? :)

I can see this trickling-in of titles extending indefinitely due to the nature of the whitelisting combined with wishing to stay responsive to new releases/updates.
Cheers.

ivyl:
The previous patch is in experimental, so you can just keep on creating new PRs targeting that. No need to that that for this commit though. I'll pick it up and close this PR.

I'm tempted to try enabling nvapi for all games with the next Proton rebase. Have you found any games that would be broken with that? Not even talking about DLSS, just any game that required nvapihack in the past.

adamnv:
I added one more, sorry.  Happy to move it to a new PR if you've already pulled the trigger on the previous batch. :)

> I'm tempted to try enabling nvapi for all games with the next Proton rebase. Have you found any games that would be broken with that? Not even talking about DLSS, just any game that required nvapihack in the past.

It's certainly tempting, though I have found some titles (three or four) that break with that.  Maybe they're fixable or already fixed at the dxvknvapi level (not sure how closely Proton tracks dxvknvapi master?) or they're just unveiling poor app assumptions but that's another investigation I think.

ivyl:
> It's certainly tempting, though I have found some titles (three or four) that break with that. Maybe they're fixable or already fixed at the dxvknvapi level

Do you happen to remember the names? Indeed, those would be interesting to look at / fix.

> not sure how closely Proton tracks dxvknvapi master?

I bump dxvk-nvapi whenever Jens asks me to do so, but it wouldn't be hard to start tracking master in experimental. We already do that for dxvk and vkd3d-proton.

adamnv:
>The previous patch is in experimental, so you can just keep on creating new PRs targeting that. No need to that that for this commit though. I'll pick it up and close this PR.

Since I had yet another batch of updates queued up, I've just rolled any outstanding updates from this PR into that one ( https://github.com/ValveSoftware/Proton/pull/6227 ), so this one can be closed.

PR #6227, opened 2022-10-12
Enable dxvknvapi on more (DLSS) titles I've verified as functioning & stable

adamnv:
Expanding the verified list introduced in https://github.com/ValveSoftware/Proton/pull/6120 .   This time, also, targeted at experimental_7.0.  Thanks.


adamnv:
I'd appreciate your thoughts on a discovery, @ivyl :
**Red Dead Redemption 2** requires `PROTON_ENABLE_NVAPI=1` and _also_ `PROTON_HIDE_NVIDIA_GPU=0` for DLSS to be correctly detected (even under its default Vulkan renderer, yeah!).

Since I don't think `PROTON_HIDE_NVIDIA_GPU=1` is the default for all apps, I assume it's being explicitly set in a game profile somewhere, but after grepping proton I can't seem to detect where; any clues?

Or maybe `PROTON_HIDE_NVIDIA_GPU=1` is actually the default for all apps and I should just explicitly add `PROTON_HIDE_NVIDIA_GPU=0` for RDR2.  Or `PROTON_HIDE_NVIDIA_GPU=0` should just be implied for titles needing the `"enablenvapi"` quirk in the proton launch script.

edit: a data point from testing which I forgot to add - just `PROTON_HIDE_NVIDIA_GPU=0` on its own without `PROTON_ENABLE_NVAPI=1` will cause a launch failure.  I'm guessing that's the reason `PROTON_HIDE_NVIDIA_GPU=1` was forced for this title before dxvknvapi was part of Proton - the title gets really confused if it thinks it detects nvidia but nvapi fails to initialize.

ivyl:
> I'd appreciate your thoughts on a discovery, @ivyl : **Red Dead Redemption 2** requires `PROTON_ENABLE_NVAPI=1` and _also_ `PROTON_HIDE_NVIDIA_GPU=0` for DLSS to be correctly detected (even under its default Vulkan renderer, yeah!).
> 
> Since I don't think `PROTON_HIDE_NVIDIA_GPU=1` is the default for all apps, I assume it's being explicitly set in a game profile somewhere, but after grepping proton I can't seem to detect where; any clues?

Ah, it's another flavor of the nvapihack we had to add.

If you look into `proton` script you can see that `PROTON_HIDE_NVIDIA_GPU` gets converted into `hidenvgpu` compat option.

Steam provides us with compat config options using [SteamPlay 2.0 Manifest](https://steamdb.info/app/891390/info/) which indeed has this option set for a few games:

```
1174180/comment | Red Dead Redemption 2
1174180/config | hidenvgpu,nativevulkanloader

997070/comment | Marvel's Avengers
997070/config | hidenvgpu

1404210/comment | Red Dead Online
1404210/config | hidenvgpu,nativevulkanloader
```

> Or maybe `PROTON_HIDE_NVIDIA_GPU=1` is actually the default for all apps and I should just explicitly add `PROTON_HIDE_NVIDIA_GPU=0` for RDR2. Or `PROTON_HIDE_NVIDIA_GPU=0` should just be implied for titles needing the `"enablenvapi"` quirk in the proton launch script.

I think `enablenvapi` would need to override `hidenvgpu`.
 
> edit: a data point from testing which I forgot to add - just `PROTON_HIDE_NVIDIA_GPU=0` on its own without `PROTON_ENABLE_NVAPI=1` will cause a launch failure. I'm guessing that's the reason `PROTON_HIDE_NVIDIA_GPU=1` was forced for this title before dxvknvapi was part of Proton - the title gets really confused if it thinks it detects nvidia but nvapi fails to initialize.

We have dxgi nvapi hack for the longest time. Even when we were using wined3d's dxgi we made changes to dxvk so all the logic and overrides are a part of dxvk_config.dll and we used that for overrides on our end.

If memory serves me well I think that RDR2 was using also some other API (Vulkan?) and didn't like either the mismatch or tried to load nvapi if it saw Nvidia's vendor id there.

adamnv:
> I think enablenvapi would need to override hidenvgpu

Thanks!  I'll go for that angle I reckon.

Bitwolfies:
The Uncharted Legacy of Thieves collection could also be added to this list in my testing, it works on all settings, no crashes.

proton
```diff
@@ -1420,6 +1422,8 @@ class Session:
 
         if "enablenvapi" in self.compat_config:
             self.env["DXVK_ENABLE_NVAPI"] = "1"
+            # enablenvapi beats hidenvgpu
+            self.env["PROTON_HIDE_NVIDIA_GPU"] = "0"
```
ivyl: Have you tested that it works? It doesn't look like it does.

in `init_session()` we have a bunch of checks in the form:

```python
self.check_environment("PROTON_HIDE_NVIDIA_GPU", "hidenvgpu")
self.check_environment("PROTON_ENABLE_NVAPI", "enablenvapi")
```

This converts the environment variables to entries in `self.compat_config`.

If you look a bit down `hidenvidiagpu` is handled as follows:

```python
if "hidenvgpu" in self.compat_config:
    self.env["WINE_HIDE_NVIDIA_GPU"] = "1" 
```

Converting the `compat_config` value in `WINE_` prefixed variable that's then interpreted by Wine.

That's a few lines above this line, so it is too late to affect the outcome.

Something like this (untested) should make more sense:

```diff
diff --git a/proton b/proton
index 47e51897..46edef0a 100755
--- a/proton
+++ b/proton
@@ -1317,6 +1317,12 @@ class Session:
         self.check_environment("PROTON_HEAP_DELAY_FREE", "heapdelayfree")
         self.check_environment("PROTON_ENABLE_NVAPI", "enablenvapi")
 
+        if "enablenvapi" in self.compat_config:
+            self.env["DXVK_ENABLE_NVAPI"] = "1"
+            # enablenvapi beats hidenvgpu
+            if "hidenvgpu" in self.compat_config:
+                self.compat_config.remove("hidenvgpu")
+
         if "noesync" in self.compat_config:
             self.env.pop("WINEESYNC", "")
         else:
```

ivyl:
I've pushed all but the RDR2 commits to experimental. It should appear in bleeding-edge soon. Thanks!

adamnv:
> Have you tested that it works? It doesn't look like it does.

Yeah well spotted!  I did test that it works, but I had a testing error (I'd accidentally left the env var in the launch options too...)

I've updated the RDR2 change again to override the right env var; would you like to pick it up from this branch or should I open a new PR?
Thanks.

Bitwolfies:
Sackboy is good too!

ivyl:
This is now merged to experimental and should appear in the next experimental release.

It's already in the bleeding-edge: https://github.com/ValveSoftware/Proton/commits/experimental-bleeding-edge-7.0-28117-20221101-pd31d72-w9c8813-d0b9f78-v4df366

I've taken a slightly different approach to overriding hidenvgpu.

Feel free to open more PRs with more games.

Issue #6277, opened 2022-10-29
Running steam in KDE sometimes makes proton profile creation fail


Patola:
# Compatibility Report
- Name of the game with compatibility issues: Earth's Shadow
- Steam AppID of the game: 1655940

## System Information
- GPU: RX 6800 XT
- Driver/LLVM version: Mesa 22.2.1
- Kernel version: 6.0.5-271-tkg-bmq
- Link to full system information report as [Gist](https://gist.github.com/): https://gist.github.com/Patola/dbfea974fdbb08617b1585a48700161d
- Proton version: 7.0-4

## I confirm:
- [X] that I haven't found an existing compatibility report for this game.
- [X] that I have checked whether there are updates for my system available.

[steam-1655940.log](https://github.com/ValveSoftware/Proton/files/9892519/steam-1655940.log)

## Symptoms 
Don't even bother trying to read the proton log, it's basically only the header -- yes, it's empty.
Launching the game with no proton compatibility version set, or any proton compatibility version set (e.g. Proton 7.0-4, Proton Experimental, even GE-Proton-7-30) only creates the directory compatdata/1655940, a zero-bytes pfx.lock and nothing else.

## Reproduction
* Set for Proton-7.0-4, before launching:
`[2:25] [6376] [patola@risadinha patola]% ls -l /jogos/steam/steamapps/compatdata/1655940`
`ls: cannot access '/jogos/steam/steamapps/compatdata/1655940': No such file or directory`
* Press play, green button turns blue and immediately turns back green. Usually (for other games) it takes a few seconds building the proton prefix for the first time.
* After play:
`[2:27] [6377] [patola@risadinha patola]% ls -l /jogos/steam/steamapps/compatdata/1655940`
`total 0`
`-rwxr-xr-x 1 patola patola 0 Oct 29 02:28 pfx.lock`
`[2:28] [6377] [patola@risadinha patola]%`
* The game files seem correct:
`[2:28] [6377] [patola@risadinha patola]% ls -la /jogos/steam/steamapps/common/Earth\'s\ Shadow`
`total 356`
`drwxr-xr-x   4 patola patola   4096 Oct 29 02:04 .`
`drwxr-xr-x 583 patola patola  20480 Oct 29 01:58 ..`
`drwxr-xr-x   4 patola patola   4096 Oct 29 02:04 EarthShadow`
`-rwxr-xr-x   1 patola patola 327680 Oct 29 01:58 EarthShadow.exe`
[...]

PR #6341, opened 2022-11-25
proton script: Don't crash if sys.stderr is not usable

smcv:
If a Steam user runs Steam from a terminal, puts it in the background and then exits from that terminal, or if they restart their desktop session from a terminal (as in ValveSoftware/Proton#6277) and then exit from that terminal, then we can inherit a stdout and/or stderr file descriptor pointing to an invalid file descriptor. Writing to such a file descriptor fails with EIO. Similarly, we could get write errors as a result of OS state, such as ENOSPC if we are writing to a disk that is full, or EPIPE if a stream to a logging framework such as the systemd journal has been shut down.

In sufficiently pathological situations, the file descriptor could even become invalid while the `proton` script is running, so even checking for validity on startup would not be enough to prevent this.

The ability to log to stderr is important but not functionally critical, and it's not like there is anything we can usefully do about a write failure here (or even anywhere we can usefully put a warning message), so just ignore write errors. This is similar to the behaviour of the `logging` framework in the Python standard library (which writes to `stderr` if a user-defined handler fails, but takes no other action) and also similar to the approach taken to solve
ValveSoftware/steam-for-linux#8069.

### DIFF
--- a/proton
+++ b/proton
@@ -68,8 +68,14 @@ def append_to_env_str(env, variable, append_str, separator):
         env[variable] = env[variable] + separator + append_str
 
 def log(msg):
-    sys.stderr.write(PFX + msg + os.linesep)
-    sys.stderr.flush()
+    try:
+        sys.stderr.write(PFX + msg + os.linesep)
+        sys.stderr.flush()
+    except OSError:
+        # e.g. see https://github.com/ValveSoftware/Proton/issues/6277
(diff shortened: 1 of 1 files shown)


ivyl:
Thanks! I've merged the change into experimental. It should appear in the bleeding edge soon and from there it will trickle to releases.

PR #6457, opened 2023-01-16
Please backport this patch from Wine 8.0 Rc4 to support incoming MinGW-w64 11.0.0

rboxeur:
The only way to build Wine for 32 bits while building a Wow Wine 64 with Git MinGW-w64 is to apply this patch
https://bugs.winehq.org/show_bug.cgi?id=54263

Can you please apply it so we could anticipate the incoming release for MinGW-w64 and rebased Proton based on Wine 8.0

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -3,7 +3,7 @@
 	url = ../wine
 [submodule "dxvk"]
 	path = dxvk
-	url = https://github.com/doitsujin/dxvk.git
+	url = ../dxvk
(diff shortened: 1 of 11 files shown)


kisak-valve:
Hello @rboxeur, this merge request is a mess and can't be used at all.

For convenience it appears that https://gitlab.winehq.org/wine/wine/-/merge_requests/1921/ is the actual code in question.

Issue #6415, opened 2022-12-26
Pentiment (1205520)


vassilmladenov:
# Compatibility Report
- Name of the game with compatibility issues: Pentiment
- Steam AppID of the game: 1205520

## System Information
- Steam Deck
- Software Update Channel: Stable
- OS Version: 3.4.2
- Proton version: 1667938887 proton-7.0-5

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

[Proton log](https://github.com/ValveSoftware/Proton/files/10304765/steam-1205520.log): **NOTE:** the web view **does not** pop up at all when running with `PROTON_LOG=1`. I can only make it come up when the logging flag is not set.

This game is listed as Deck Verified.

## Symptoms

Log into Xbox Network (syncs achievements/etc with Xbox Live) does not work. It will variously lock up the game in gaming mode or open a web view that eventually gets stuck on a blank white screen in desktop mode. When running with `PROTON_LOG=1`, the web view does not come up at all, which is interesting.

Miscellaneous other issues: I cannot uncheck Full Screen in the settings, and the display resolution dropdown is blank.

## Reproduction

1) Launch game in gaming mode
2) Click Log into Xbox Network (Y)
3) Game gets stuck "Logging In..."
4) Force the game to quit with Steam button
5) Switch to desktop mode (attach keyboard and mouse)
6) Launch game
7) Click Log into Xbox Network with mouse
![Screenshot_20221226_115555](https://user-images.githubusercontent.com/5396637/209581445-79b9fbb1-421c-449b-9881-ea3e60ad81a6.png)
8) Use keyboard Super key to get the taskbar focused
9) Minimize game from the right click menu of the taskbar icon
10) Enter credentials for Microsoft account in the web view - NOTE: I needed to use a different Windows PC to approve the app the first time.
11) Web view gets stuck on white screen instead of logging in the game
![Screenshot_20221226_115416](https://user-images.githubusercontent.com/5396637/209581430-98465673-02f5-4063-9685-8710fbee7a86.png)

tgurr:
## System Information
- GPU: NVIDIA GeForce RTX 2060 SUPER
- Driver/LLVM version: nvidia 525.78.01
- Kernel version: 6.1.3
- Link to full system information report as [Gist](https://gist.github.com/): https://gist.github.com/tgurr/be1aee876e51942bf1de10c6ec8a8409
[...]

PR #6479, opened 2023-01-24
proton: Add workaround for Pentiment to mimic Steam Deck depot fix.

tgurr:
The game works out of the box on Steam Deck since there's a separate depot which ships a 0 byte SpeechSynthesisWrapper.dll file, this fix however is missing on regular Linux environments where the game fails to run.

https://steamdb.info/depot/1205522/
https://github.com/ValveSoftware/Proton/issues/6415#issuecomment-1379357568

### DIFF
--- a/proton
+++ b/proton
@@ -1498,6 +1498,14 @@ class Session:
         else:
             argv = [g_proton.wine64_bin, "c:\\windows\\system32\\steam.exe"]
 
+        # Pentiment workaround to also work on regular Linux (mimic Steam Deck depot behaviour)
+        # https://steamdb.info/depot/1205522/
+        if os.environ.get("SteamGameId", 0) == "1205520":
+            pentiment_gameplugins_dir = os.environ['STEAM_COMPAT_INSTALL_PATH'] + '/Pentiment_Data/Plugins/x86_64/'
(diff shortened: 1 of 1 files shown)


Bitwolfies:
Question, does the removal of this DLL actually effect anything? Seems odd they'd ship it on windows if it supposedly does nothing.

Not sure if I like the idea of devs shipping depots with gimped functionality cause something doesn't yet work in proton, since you know they'll never go back these games and remove the workarounds.

tgurr:
See the mentioned [comment](https://github.com/ValveSoftware/Proton/issues/6415#issuecomment-1379357568) from @bklooster already linked above for clarification, he seems to have some quite deep insights on that and might be the right person to talk to in this regards. My strong guess is that it is required for the Text-to-Speech "Voice Assist" feature the game offers as an accessibility feature, for which on Linux the menu only offers an "Off" setting for that option.

Edit: If you implement a/the suggested fix in proton itself this would allow to get rid of the extra depot as well and it could be easily reverted when the missing function is implemented in proton/wine.

ivyl:
`WINEDLLOVERRIDES=SpeechSynthesisWrapper.dll=d %command%` should be enough to help here. If that's the case a dll override in `wine.inf.in` is better.

ivyl:
I've merged the `wine.inf` version of this hack.

PR #6555, opened 2023-02-21
steam_helper: separate parameters from command when using ShellExecuteW

Jan200101:
ShellExecuteW ignores any extra information passed as part of lpFile
resulting in arguments being ignored.

### DIFF
--- a/steam_helper/steam.cpp
+++ b/steam_helper/steam.cpp
@@ -1028,21 +1028,32 @@ static BOOL streq_niw(const WCHAR *l, const WCHAR *r, size_t len)
     return TRUE;
 }
 
-static BOOL should_use_shell_execute(const WCHAR *cmdline)
+static WCHAR* get_end_of_excutable_name(WCHAR *cmdline)
 {
-    BOOL use_shell_execute = TRUE;
     BOOL quoted = FALSE;
-    const WCHAR *executable_name_end = cmdline;
+    WCHAR *executable_name_end = cmdline;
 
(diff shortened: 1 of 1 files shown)


GloriousEggroll:
been testing this for a while, unfortunately it breaks a few games:

Guilty Gear Xrd
Escape from Monkey Island
Memento Mori

Jan200101:
rebased onto 8.0 and fixed the issue mentioned by GloriousEggroll tested using Guilty Gear Xrd REV 2 and Ultrakill to ensure nothing else broke.

To go into more issue what caused this issue:
GGXrd uses a bat to launch which is started using an absolute path by Steam
steam_helper converts that absolute unix path into an absolute dos path and quotes it
the logic within `get_end_of_excutable_name` respects this and stops iterating the cmdline once it finds the end of quote
This leaves our pointer dangling at the end quote instead of after the program

I updated the `get_end_of_excutable_name` function to go to the full end, quotes included, and checking for a quote within `should_use_shell_execute`.

GloriousEggroll:
Confirming everything working as expected now, thanks @Jan200101

ishitatsuyuki:
@ivyl This seems ready for review, and according to the user is necessary for a Titanfall 2 mod (Northstar) to work. Would you mind taking a look?

ivyl:
https://github.com/ValveSoftware/Proton/commit/6b9be3305e43f212216cf0a6fe6650d953911880 merged, thanks!

PR #6633, opened 2023-03-21
proton: Fix a couple simple crashes

alice945:
Fixes a crash when no verb is provided due to accessing a nonexistent index in `sys.argv`.
Allows the `STEAM_COMPAT_CLIENT_INSTALL_PATH` environment variable to not be set.

### DIFF
--- a/proton
+++ b/proton
@@ -792,7 +792,7 @@ class CompatData:
                 os.symlink("/", self.prefix_dir + "/dosdevices/z:")
 
             # collect configuration info
-            steamdir = os.environ["STEAM_COMPAT_CLIENT_INSTALL_PATH"]
+            steamdir = os.environ.get("STEAM_COMPAT_CLIENT_INSTALL_PATH", "")
 
             use_wined3d = "wined3d" in g_session.compat_config
             use_dxvk_dxgi = not use_wined3d and \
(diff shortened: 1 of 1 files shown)


proton
```diff
@@ -792,7 +792,7 @@ class CompatData:
                 os.symlink("/", self.prefix_dir + "/dosdevices/z:")
 
             # collect configuration info
-            steamdir = os.environ["STEAM_COMPAT_CLIENT_INSTALL_PATH"]
+            steamdir = os.environ.get("STEAM_COMPAT_CLIENT_INSTALL_PATH", "")
```
ivyl, line 795: We depend on Steam setting up this variable and using `[]` will make this fail in a clear way if this ever regresses in Steam.

By doing `.get()` the way you do here we would silence this but also copying of the steam DLLs would silently fail.

ivyl:
I assume this is an attempt to make Proton usable outside of Steam? I think it's a good thing overall. IMO we should reorganize code a bit:

1. Don't fail if environment variables / verb are not provided.
2. Print a help / refer to README.
3. The current verbs should still encode assumptions about the environment and fail in an obvious way.
4. Introduce new verbs that allow you to run outside of the Steam Client.

PR #6673, opened 2023-04-08
Remove LLVM from issue report templates

kisak-valve:
The amount of games influenced by LLVM version is fairly low these days. Low priority adjustment when it is convenient.

EDIT: Also ask users to also share Steam Runtime Information now that has been split from Steam's system information window.


ivyl:
I've merged it on top of pending update to Proton 8.0-3. Once this version goes live the default branch of the repo (`proton_8.0`) will get updated and include the template changes. Thanks!

PR #6911, opened 2023-07-04
Reduce media-converter GStreamer rank

Alcaro:
The avdec_wmav1 filter has rank Marginal; if the user has installed that, it should be used instead of media-converter


Alcaro:
Seems that decodebin ignores elements with rank below MARGINAL. Sorry about that, should've tested it better.

decodebin ranks things alphabetically as a tiebreaker, so setting ranks to MARGINAL and names to zzz_protonmediaconverter and zz_protonmediaconverterbin would yield the results I want.

But that's just way too ugly. Better tell people to uninstall mediaconverter.

(also, what the h_h is going on with the commit list)

Issue #7108, opened 2023-09-17
--steam-runtime option in configure.sh is not used


nm004:
`--steam-runtime` option is not used in `cofigure.sh` script, but the help message mentions that option.

```
./configure.sh --help
::   Steam Runtime
::     Proton builds that are to be installed & run under the steam client must be built with
::     the Steam Runtime SDK to ensure compatibility.  See README.md for more information.
:: 
...
::     --steam-runtime=soldier  Name of the steam runtime release to build for (soldier, scout).
...
```

### Steps to reproduce

Issue the following command.

```
./configure.sh --steam-runtime=sniper
```

### Actual Result

```
!! Unrecognized option --steam-runtime
!! Usage: ../configure.sh { --no-proton-sdk | --proton-sdk-image=<image> --steam-runtime=<name> }
...
```

### Expected Result

Should we delete the help text of that option?

ivyl:
The PR was merged into bleeding-edge. Thank you.

PR #7114, opened 2023-09-21
configure.sh: drop --steam-runtime help text

nm004:
This commit dropped the help text for --steam-runtime option. That option is no longer used since the commit bc63dbbf2a041b.

This commit resolves issue #7108.


ivyl:
merged into bleeding-edge, thanks!

PR #7138, opened 2023-09-27
Refresh request for Steam Runtime information

kisak-valve:
Looks like this menu option got renamed in the Steam client.


oliverklee (approved):

ivyl:
It's in bleeding edge and will trickle to other branches later on.

PR #7253, opened 2023-11-11
proton: Enable amdags for The Last of Us Part I

tgurr:
`WINEDLLOVERRIDES="amd_ags_x64=b"` was required before to fix:

* outdated graphics driver warning on game start
* wrong vram reported in options
* delayed texture loading

https://github.com/ValveSoftware/Proton/issues/6653
https://www.protondb.com/app/1888930

Tested with `PROTON_ENABLE_AMD_AGS=1` on mesa 23.3.0-rc3 with a RX7800XT on current Proton Experimental `experimental-8.0-20231107c`.

### DIFF
--- a/proton
+++ b/proton
@@ -1225,6 +1225,7 @@ def default_compat_config():
         if appid in [
                 "1245620", #Elden Ring
                 "1888160", #Armored Core VI
+                "1888930", #the last of us part 1
                 "814380", #Sekiro: Shadows Die Twice
                 "2379390", #Rainbow Six Extraction
                 ]:


alasky17:
Thank you for the PR!  Unfortunately setting amd_ags_x64=b also causes a performance drop on at least some hardware.  Our goal is to set amd_ags to built-in for TLOU for the benefits you mentioned, but I think it is a bit premature to set this as the default in Proton until we solve the perf drop issue.  Folks can set this manually in the meantime if they are not seeing (or not bothered by) the perf drop.  Thanks again for taking the time to submit this!

tgurr:
Interesting, thanks for the clarification, details and response! For me this was actually a performance boost as the game couldn't use the available vram so my guess was this would apply to pretty much any usecase and also having seen the newly introduced proton `enableamdags` got applied to many other games lately. Afaik there is/was also a hotfix proton version, not sure what that actually does. I wasn't aware of performance issues/dops on other/some hardware. Hope this can or maybe has already been sorted out in the meantime due to updates and also ags code changes and this could be enabled some time in the future to get a nice out of the box experience. My hardware testing capabilities are quite limited so I'm not sure I'm able to contribute more in this regard and you guys obviously know much better. Feel free to leave the PR open until you think the issues are solved so it won't get forgotten or just close it to your liking.

alasky17:
It was helpful that this game got brought up in the context of switching to amd_ags_x64=b because we are indeed working on switching games over as you noticed :)  So fixing the perf issue that some users encountered has been helpful towards that goal (with the additional benefit of improving the out-of-the-box experience for users with similar hardware to you it seems :D).  If you or anyone else reading this has found any bugs that only happen when you enable amd_ags_x64=b, it would be very helpful if you could tag me on the github issue for the game so that we can take a look :)

alasky17:
@tgurr We looked into the performance drop (which as I mentioned earlier we wanted to fix anyways) and @gofman fixed it :D So I went ahead and cherry-picked your PR and it should be in the bleeding-edge beta branch of experimental soon :)

PR #7335, opened 2023-12-11
Change DXVK-NVAPI to enabled-by-default

adamnv:
I've tested this with Guardians of the Galaxy (i.e. it auto-disables dxvknvapi correctly), and Ratchet & Clank (i.e. it default-enabled dxvknvapi correctly) and a light smattering of other titles so far.


proton
```diff
@@ -1461,8 +1329,8 @@ class Session:
             if not "VKD3D_FEATURE_LEVEL" in self.env:
                 self.env["VKD3D_FEATURE_LEVEL"] = "12_0"
 
-        # enablenvapi beats hidenvgpu
-        if "hidenvgpu" in self.compat_config and "enablenvapi" not in self.compat_config:
+        # hidenvgpu requires disablenvapi to also have been specified
```
adamnv, line 1332: I think hidenvgpu should probably *imply* disablenvapi now, but I didn't want to change the logic in what's otherwise this straightforward enablenvapi->disablenvapi inversion PR.

ivyl: It may be fine, if we are hiding the gpu vendor nvapi should do no harm. I'll make sure we test that.

Issue #3346, opened 2019-12-22
Descent 3 (273590)


NathanHannon:
Proton 4.11-11
Descent 3 Crashes during the intro movie. The dump indicated a memory access violation with the following:
Call stack:
------------------------------------------------------------------
    main.exe 0056fa7c()
    kernel32.dll 7b47faf2()
    kernel32.dll 7b481cec()
    kernel32.dll 7b47fafe()
------------------------------------------------------------------

kisak-valve:
Hello @NathanHannon, please add [`PROTON_LOG=1 %command%`](https://github.com/ValveSoftware/Proton#runtime-config-options) to the game's launch options and drag and drop the generated $HOME/steam-$APPID.log into the comment box. Also, please copy your system information from Steam (`Steam` -> `Help` -> `System Information`) and put it in a [gist](https://gist.github.com/), then include a link to the gist in this issue report.

NathanHannon:
System Info: https://gist.github.com/NathanHannon/6d8aa2d0ccccf8c3a9c6fcfc7fdd9850
Log Output: 
[steam-273590.log](https://github.com/ValveSoftware/Proton/files/3992134/steam-273590.log)

PR #7365, opened 2023-12-21
proton: Enable gamedrive compat option for Descent 3

Frogging101:
Descent 3 corrupts its stack if the installation path is too long. The return pointer gets overwritten with the path to the intro movie and we crash.

Ref #3346


ivyl:
pushed to experimental, should show up in bleeding-edge soon, thanks!

PR #7424, opened 2024-01-19
wineopenxr: change instance type to session type

Supreeeme:
Currently, wineopenxr determines the graphics backend by the instance extensions passed to it, but there's nothing stopping an application from enabling multiple graphics extensions. If an application passes multiple, proton would incorrectly assume all sessions will use the backend from the last instance extension passed in. This behavior can be observed when using a Windows build of [OpenComposite](https://gitlab.com/znixian/OpenOVR) with any OpenVR game, with an OpenComposite build that has more than one backend enabled, which the prebuilt dlls have by default. This PR fixes that. OpenXR graphics backends are also handled on a per session basis anyway and not per instance, so this just makes more sense.

### DIFF
--- a/proton
+++ b/proton
@@ -1746,6 +1746,9 @@ if __name__ == "__main__":
         log("No compat data path?")
         sys.exit(1)
 
+    # CW-Bug-Id: #23185 Avoid triggering new SDL 2.30 behavior.
+    os.environ.pop('SteamVirtualGamepadInfo', None)
+
     g_proton = Proton(os.path.dirname(sys.argv[0]))
 
     g_proton.cleanup_legacy_dist()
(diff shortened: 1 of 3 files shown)


gofman:
Looks good to me in general, a couple of nits.

- variable 'type' is now unused in wine_xrCreateInstance(), please remove it;

```
> @@ -1007,6 +1000,7 @@ XrResult WINAPI wine_xrCreateSession(XrInstance instance, const XrSessionCreateI
>     }
> 
>     wine_session->wine_instance = wine_instance;
> +    wine_session->session_type = session_type;
```

session_type might be uninitialized. Actually, it would be better if we would just return an error from wine_xrCreateSession on unsupported binding, but it is unrelated to this patch and shouldn't be squashed in. I guess, just ```uint32_t session_type = 0;```

Supreeeme:
@gofman Done

gofman:
This is now in Proton Experimental [bleeding-edge]. Thanks for the patch!

PR #7555, opened 2024-03-06
proton: Add Helldivers 2 to MFDXGI hack

anone-zone:
Works around black screen issue if launching the game in fullscreen mode.

See https://github.com/ValveSoftware/Proton/issues/7486


ivyl:
Hi. The issue looks intermittent. Repro is not 100% reliable either with MFDXGI hack or without. The repro rate also doesn't seem to change.

We'll have a look at the black screen issue.

PR #7876, opened 2024-07-05
proton: Disable fsync and esync for Flatout 3.

doZennn:


doZennn:
d47668908edc0c0d9cf56b48828261645a11caaa

Issue #7905, opened 2024-07-15
duplicate call to `xrCreateInstance` when using OpenComposite + Monado


emily-is-my-username:
Currently, VR initialization in Proton seems to *require* OpenVR and thus SteamVR, even for OpenXR-only games, which complicates setup for other OpenXR runtimes such as Monado. One workaround ([as suggested here](https://github.com/ValveSoftware/Proton/issues/6038#issuecomment-1590246971)) is using OpenComposite to bridge all OpenVR calls to the OpenXR runtime.

However, if OpenComposite uses the same openxr_loader library as Proton, VR initialization still fails with: ```Loader does not support simultaneous XrInstances```. There are two OpenXR instances active simultaneously, which is unsupported by the loader.

As @eburema explained [here](https://github.com/ValveSoftware/Proton/issues/7382#issuecomment-1973353179) and [in the OpenComposite issue](https://gitlab.com/znixian/OpenOVR/-/issues/416#note_1963119445), that happens because of the order of operations in `steam_helper`'s `initialize_vr_data`:  First, OpenVR is initialized, and the first OpenXR instance is created through OpenComposite, then `wineopenxr`'s `__wineopenxr_get_extensions_internal` is called, tries to create the second instance and triggers the error.

This is easily fixed by cleaning up the OpenVR client before invoking the OpenXR loader, as the client is not used anymore and gets closed later in the function anyway.

There may be better ways to support other OpenXR runtimes, for example by requesting the XR extensions independently from OpenVR - directly from the OpenXR runtime (as suggested [here](https://github.com/ValveSoftware/Proton/issues/6038)), but I deliberately want to keep the scope as small as possible - this is **not** intended as a feature request for out-of-the-box Monado compatibility.

I also understand that this is not strictly an issue with Proton itself, but using OpenComposite is a useful workaround to run OpenXR games in Monado without changing the OpenVR / SteamVR specific parts of Proton.

---

OpenComposite issue:
- https://gitlab.com/znixian/OpenOVR/-/issues/416

Related Proton issues:
- https://github.com/ValveSoftware/Proton/issues/7382 (error in log seems to be a direct result of this bug)
- https://github.com/ValveSoftware/Proton/issues/6038 (general feature request for Monado compatibility)

Fix:
- https://github.com/ValveSoftware/Proton/pull/8126

PR #7906, opened 2024-07-15
steam_helper: get OpenXR extensions before initializing OpenVR

emily-is-my-username:
When using OpenComposite, both OpenVR and OpenXR functions will call the same underlying OpenXR loader.

Because the OpenXR loader only supports a single active instance, `initialize_vr_data` currently fails because an OpenXR instance has already been initialized when XR extensions are queried.

For SteamVR, the runtime already needs to be running for `xrCreateInstance` to succeed.

To support both setups, this commit tries to query XR extensions before OpenVR initialization, but still defaults to the previous behavior if the first try was unsuccessful.

Fix for:
https://github.com/ValveSoftware/Proton/issues/7905

### DIFF
--- a/steam_helper/steam.cpp
+++ b/steam_helper/steam.cpp
@@ -684,9 +684,66 @@ static void *get_winevulkan_unix_lib_handle(HMODULE hvulkan)
     return dlopen(info.dli_fname, RTLD_NOW);
 }
 
+
+static int WINAPI save_openxr_extensions_to_registry(HKEY vr_key){
+    int (WINAPI *p__wineopenxr_get_extensions_internal)(char **instance_extensions, char **device_extensions, uint32_t *physdev_vid, uint32_t *physdev_pid);
+    char *xr_inst_ext, *xr_dev_ext;
+    HMODULE hwineopenxr;
+    LSTATUS status;
+
+    if ((hwineopenxr = LoadLibraryA("wineopenxr.dll")))
+    {
+        p__wineopenxr_get_extensions_internal = reinterpret_cast<decltype(p__wineopenxr_get_extensions_internal)>
+                (GetProcAddress(hwineopenxr, "__wineopenxr_get_extensions_internal"));
+        if (p__wineopenxr_get_extensions_internal)
+        {
+            uint32_t vid, pid;
+            if (!p__wineopenxr_get_extensions_internal(&xr_inst_ext, &xr_dev_ext, &vid, &pid))
+            {
+                WINE_TRACE("Got XR extensions.\n");
+                if ((status = RegSetValueExA(vr_key, "openxr_vulkan_instance_extensions", 0, REG_SZ,
+                        (BYTE *)xr_inst_ext, strlen(xr_inst_ext) + 1)))
+                {
+                    WINE_ERR("Could not set openxr_vulkan_instance_extensions value, status %#x.\n", status);
+                    return STATUS_UNSUCCESSFUL;
+                }
+                if ((status = RegSetValueExA(vr_key, "openxr_vulkan_device_extensions", 0, REG_SZ,
+                        (BYTE *)xr_dev_ext, strlen(xr_dev_ext) + 1)))
+                {
+                    WINE_ERR("Could not set openxr_vulkan_device_extensions value, status %#x.\n", status);
+                    return STATUS_UNSUCCESSFUL;
+                }
+                if ((status = RegSetValueExA(vr_key, "openxr_vulkan_device_vid", 0, REG_DWORD,
+                        (BYTE *)&vid, sizeof(vid))))
+                {
+                    WINE_ERR("Could not set openxr_vulkan_device_vid value, status %#x.\n", status);
+                    return STATUS_UNSUCCESSFUL;
+                }
+                if ((status = RegSetValueExA(vr_key, "openxr_vulkan_device_pid", 0, REG_DWORD,
+                        (BYTE *)&pid, sizeof(pid))))
+                {
+                    WINE_ERR("Could not set openxr_vulkan_device_pid value, status %#x.\n", status);
+                    return STATUS_UNSUCCESSFUL;
+                }
+            }
+        }
+        else
+        {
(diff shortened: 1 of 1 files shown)


gofman:
Thanks for the patch!

> For SteamVR, the runtime already needs to be running for xrCreateInstance to succeed.

Did you check that? It would be very weird, I am not aware there ever was such a bug.

As I understood the description here, the problem with OpenComposite is only when VR and XR instances are open at the same time and the order of opening doesn't matter. Is it correct? If it is, AFAIU the only problem is that VR instance is closed in the end of the function, does just closing it after VR queries are done before dealing with XR help? If yes, we may probably try to go this way after some smoke test with SteamVR. If it doesn't for some reason, would you please elaborate more on the actual issue with OpenComposite.

Broadly, I think it should be fixed in OpenComposite. There are games which may be doing whatever with initialization.  If workint that around that is a matter of a small and sensible change (like release VR instance earlier) than its fine. While we probably don't want to introduce complications like in the current form of the patch for this purpose.

emily-is-my-username:
> > For SteamVR, the runtime already needs to be running for xrCreateInstance to succeed.
>
> Did you check that? It would be very weird, I am not aware there ever was such a bug.

That's what I observed on my machine with my [original proof of concept](https://github.com/emily-is-my-username/Proton/commit/8a27a536190cd3e6f2b779157ed4300b74d564ee) (didn't keep any logs though).

But you are right, that *would* be weird, I haven't really thought about it thoroughly, just assumed from [this comment](https://github.com/ValveSoftware/Proton/blob/b6d5618b6da9c90649590fedea70b0af8ba2e7e6/wineopenxr/openxr.c#L322) that (Linux) SteamVR was not intended to be started by the OpenXR loader.

Unfortunately, (my) SteamVR feels very unstable in general for a variety of reasons, and I am not sure how much of that is my / NixOS' fault, Linux compatibility in general, Proton, or actual bugs within SteamVR. I don't have a Windows machine to test what SteamVR is *supposed* to do, but I could install Ubuntu for testing, it seems to be the most supported distro most of the time.

> As I understood the description here, the problem with OpenComposite is only when VR and XR instances are open at the same time and the order of opening doesn't matter. Is it correct?
> does just closing it after VR queries are done before dealing with XR help?

Yes, AFAIK that is correct. I changed the order because I was saw that the XR part could be moved and I didn't have to change anything related to OpenVR that way..

> Broadly, I think it should be fixed in OpenComposite.

Agreed. AFAIK the VR/XR double-initialization is not a common problem outside of Proton, but i see that every game could try that and OpenComposite should generally expect it.

> [...] we probably don't want to introduce complications like in the current form of the patch for this purpose.

Also agreed. It's a niche use case and I'm not confident in my ability to test the *main* use case, so: the less changes the better :)

> If workint that around that is a matter of a small and sensible change (like release VR instance earlier) than its fine

Thank you for your input. I may be able to write a new patch next week.

eburema:
> Thanks for the patch!
> 
> > For SteamVR, the runtime already needs to be running for xrCreateInstance to succeed.
> 
> Did you check that? It would be very weird, I am not aware there ever was such a bug.
> 

It says so right here (at least if I understand the comment correctly)
https://github.com/ValveSoftware/Proton/blob/8640a5cf1960f4f76e0595c864565bc1a7b2f6f2/wineopenxr/openxr.c#L322

> As I understood the description here, the problem with OpenComposite is only when VR and XR instances are open at the same time and the order of opening doesn't matter. Is it correct? If it is, AFAIU the only problem is that VR instance is closed in the end of the function, does just closing it after VR queries are done before dealing with XR help? If yes, we may probably try to go this way after some smoke test with SteamVR. If it doesn't for some reason, would you please elaborate more on the actual issue with OpenComposite.
> 

AFAICT this is correct, sadly never was able to test it myself due to unable to get proton compiled on my system (complained about fonts or something related to that IIRC)

> Broadly, I think it should be fixed in OpenComposite. There are games which may be doing whatever with initialization. If workint that around that is a matter of a small and sensible change (like release VR instance earlier) than its fine. While we probably don't want to introduce complications like in the current form of the patch for this purpose.

If this could be easily fixed on OpenComposite side I agree since then it would also be fixed for all proton versions, though I don't know of an easy fix since OpenComposite is since it is just a library that gets loaded into wine process space and then calls OpenXR, if somehow OpenComposite could detect that it was no longer used it could release the OpenXR instance it could be made to work. (that said I just know enough about how this all works to shoot me in the foot so I might be missing something here)

Another option might to ask the folks at Krhonos group to reintroduce the ability to make multiple xrCreateInstance calls from the same process, though I think they removed that for good reasons.

gofman:
I wouldn't be opposed to a patch which would release VR instance earlier before XR initialization, but someone has to check that it actually helps anything first  (we don't test against OpenComposite in Proton development now), ideally tested with SteamVR at once to check that nothing is immediately broken. I am afraid it is not possible to make a reasonably verified patch without even compiling Proton.

eburema:
> I wouldn't be opposed to a patch which would release VR instance earlier before XR initialization, but someone has to check that it actually helps anything first (we don't test against OpenComposite in Proton development now), ideally tested with SteamVR at once to check that nothing is immediately broken. I am afraid it is not possible to make a reasonably verified patch without even compiling Proton.

Agreed it is the reason I didn't provide a patch myself since I was never able to test it, and though it would be awesome if proton could be more regularly tested against alternative implementations I completely understand that has no priority. I'm pretty sure @emily-is-my-username is able to compile and test (if there is a compiled version I can test that, both vs steamvr and opencomposite/monado, if time permits I might see about digging in why it didn't compile here)

gofman:
A side note, once there are no concerns  for the patch itself it should have real author name, surname and email on it to be pulled as is. If the author wants to stay incognito the patch might still be used but I'll have to commit it under my name referencing the auhtor's nickname and this PR in the commit message.

UPDATE By real name I don't mean exactly legal name or something formal, just what the author considers their real name, similar to, e. g., what is assumed Wine upstream.

emily-is-my-username:
closing this on in favor of https://github.com/ValveSoftware/Proton/pull/8126

PR #8005, opened 2024-08-21
Enable nativevulkanloader for Forts

tristanjl:
Fix for Forts - when running Forts with vulkan renderer enabled, everything renders transparent. Enabling "nativevulkanloader" avoids this issue.


tristanjl:
Closed - was not actually fixing the issue but just causing vulkan to fail to load and falling back to opengl

PR #8087, opened 2024-09-10
proton: Populate NVIDIA_WINE_DLL_DIR when available

liam-middlebrook:
NVIDIA-software may check the environment variable `NVIDIA_WINE_DLL_DIR` for a path where DLLs provided by the NVIDIA Linux driver can be found. This needs to be set by `proton` script, as it is ran inside of the pressure-vessel environment, so the paths returned will be valid within the environment.

Tested with an application that reads `NVIDIA_WINE_DLL_DIR` from the environment and tries to load a known DLL from it. As expected manually setting the environment variable to the on-disk location did not work, as the path is not accurate within the pressure-vessel environment.

### DIFF
--- a/proton
+++ b/proton
@@ -997,11 +997,14 @@ class CompatData:
 
                 # Try to detect known DLLs that ship with the NVIDIA Linux Driver
                 # and add them into the prefix
-                nvidia_wine_dll_dir = find_nvidia_wine_dll_dir()
-                if nvidia_wine_dll_dir:
+                if g_session.nvidia_wine_dll_dir:
                     for dll in ["_nvngx.dll", "nvngx.dll"]:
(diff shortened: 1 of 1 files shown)


adamdmoss (approved):
For what my ex-nvidia-employee opinion is worth: LGTM.  I infer from the lack of specifics (and not privileged knowledge) that this is to enable some future tech work, but if there is an existing title which needs this then pointing it out would be good for verification purposes.

liam-middlebrook:
Noting from a side-discussion between myself @ivyl, pgofman, and zf. This environment variable isn't ideal because it creates wine/proton specific behavior. Since the way this would be read on Windows isn't currently supported in Wine, this environment variable will only be read as a fallback. That way, in the future if the primary path is supported, this logic could be removed.

In the meantime there's some cleanup things to do here (like avoiding running through the C_DLL logic twice).

PR #8126, opened 2024-09-29
steam_helper: clean up OpenVR before getting OpenXR extensions in `initialize_vr_data`

emily-is-my-username:
When using OpenComposite, both OpenVR and OpenXR functions may call the same underlying OpenXR loader.

Because the OpenXR loader only supports a single active instance, `initialize_vr_data` currently fails as an OpenXR instance has already been initialized at the time XR extensions are queried.

This commit fixes the problem by cleaning up the temporary OpenVR context *before* initializing OpenXR instead of keeping it open until the end of the call.

Fix for:
https://github.com/ValveSoftware/Proton/issues/7905

----

This is a follow-up to https://github.com/ValveSoftware/Proton/pull/7906.

As @gofman [suggested](https://github.com/ValveSoftware/Proton/pull/7906#issuecomment-2261021371), it is much simpler and safer to just close the OpenVR instance early than what I did before.

Tested with:
- VRChat (OpenVR) + Proton + SteamVR
- Beat Saber (OpenXR) + Proton + SteamVR
- Beat Saber (OpenXR) + Proton + OpenComposite + Monado (this one was broken before)

### DIFF
--- a/steam_helper/steam.cpp
+++ b/steam_helper/steam.cpp
@@ -893,6 +893,11 @@ static DWORD WINAPI initialize_vr_data(void *arg)
         }
     }
 
+    if (vr_initialized) {
+        client_core->Cleanup();
+        vr_initialized = FALSE;
+    }
+
     if ((hwineopenxr = LoadLibraryA("wineopenxr.dll")))
     {
         p__wineopenxr_get_extensions_internal = reinterpret_cast<decltype(p__wineopenxr_get_extensions_internal)>


emily-is-my-username:
@gofman

I'm sorry for the long delay, but I finally found some time to try out your suggestion. It works as expected AFAICT, thank you for the guidance.

Regarding [your comment about real names](https://github.com/ValveSoftware/Proton/pull/7906#issuecomment-2261105199): I have no intention to use my real, full name on this account and I also don't want to pretend that I do :)

If your offer to pull this under your name still stands, you have my permission to do so. That goes for anyone who would be willing to do so.

emily-is-my-username:
@eburema were you able to compile proton yet? On my machine, the docker/podman build worked fine as long as my git repo and build dir were fresh. The problems I had were always caused by:
- having an unclean build directory (run `make clean`)
- having unclean git submodules (especially nested ones) after switching branches / commits (run `git submodule update --init --recursive`, verify with `git status`)

gofman:
Thanks! I pushed the patch, it is in Experimental [bleeding-edge] now: https://github.com/ValveSoftware/Proton/commit/5b26b6a1d4df17642c68ae883096a865fdc70bc6

eburema:
Finally managed to compile proton (was in the end a user error), and tested this new addition to bleeding edge with both OpenComposite + Monado, and SteamVR. Patch seems to be working fine on my end (OpenXR games started to work in OC + Monado, and nothing broke in SteamVR)

PR #8258, opened 2024-11-20
proton: fix S.T.A.L.K.E.R 2 Heart of Chornobyl

Samsuper12:
Enabled: nohardwarescheduling, heapzeromemory, heapdelayfree

### DIFF
--- a/proton
+++ b/proton
@@ -1196,6 +1196,7 @@ def default_compat_config():
                 "1404090", #Trivia Tricks
                 "2052410", #WITCH ON THE HOLY NIGHT
                 "789910", #Planet of the Apes: Last Frontier
+                "1643320", #S.T.A.L.K.E.R 2 Heart of Chornobyl
                 ]:
             ret.add("heapdelayfree")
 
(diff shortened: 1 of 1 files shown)


Samsuper12:
@ivyl

Samsuper12:
Anyway, that's not enough for loading a new game. Only for startup (loading screen, menu, etc.).

And yeah, something wrong with sound. It glitching like a sound thread has some troubles.

Continuing.

ivyl:
The game seems to be working out of the box with both AMD and Nvidia GPUs. None of the extra options seems to be required. Am I missing something?

Samsuper12:
@ivyl It seems it's occurring only on Intel ARC

ivyl:
`nohardwarescheduling` has no effect outside of Nvidia GPUs running the proprietary drivers, see [x11drv_main.c:1447](https://github.com/ValveSoftware/wine/blob/22a8cffe717b21adeef3df60cbe3af9c15e71479/dlls/winex11.drv/x11drv_main.c#L1447).

`heapzeromemory` and `heapdelayfree` also do not make much sense unless Intel's SDK started integrating code that's extremely broken. We never had this exact issue with it in the past.

Are you sure that the game is not just inherently unstable on Intel ARC? I don't think throwing some random tunables at the wall and hoping that one of them will stick is a way to go here. Crash logs would be more helpful in troubleshooting. With Intel GPUs often there's work needed in either Mesa or one of the DirectX to Vulkan layers.

Samsuper12:
Yeah, you're correct. 

I’ve just opened an issue in Mesa, which includes all the logs and my thoughts about this bug. This PR can now be closed. Thank you for your attention, and have a great day!

https://gitlab.freedesktop.org/mesa/mesa/-/issues/12185

Issue #2249, opened 2019-01-21
Source 2007 mods: Unable to join/create servers, or view achievements (218)


ghost:
# Compatibility Report
- Name of the game with compatibility issues: Source SDK Base 2007 / Source Mods
- Steam AppID of the game: 218

## System Information
- GPU: Radeon RX 560
- Driver/LLVM version: Mesa 18.0.5/6.0.0
- Kernel version: 4.15.0-43-generic
- Link to full system information report as [Gist](https://gist.github.com/): https://gist.github.com/scqless/11cf687d4980f2578ad691f56e9dda5c
- Proton version: 3.7, 3.7 Beta, 3.16, 3.16 Beta

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

[steam-9856914317583057114.log](https://github.com/ValveSoftware/Proton/files/2777227/steam-9856914317583057114.log)

## Symptoms 

Hello,

Thank you for recently adding the option to force apps to use Proton. This makes it possible to launch Source 2007 mods on Steam for Linux. However, multiplayer mods are unable to create/join servers, or view achievements. The following were observed when playing a multiplayer mod:

Server Browser: `Steam Beta must be running to make use of Find Servers`

Connect by IP: `The server requires that you be running Steam.`

Create listen server: `CD Key authentication invalid for internet servers.`

View achievements: `You must be logged in to Steam to unlock or view achievements`

## Reproduction

1. Opt into the Steam Beta
2. Force Source SDK Base 2007 to use Proton
3. Install Source SDK Base 2007
4. Install a multiplayer Source mod. I used GoldenEye: Source
5. Attempt any of the above tasks under Symptoms

nsivov:
It probably needs older API that's not exposed currently:

`12137.073:0008:0009:err:steamclient:create_win_interface Don't recognize interface name: SteamUser008`
...
`12145.249:0008:0009:err:steamclient:create_win_interface Don't recognize interface name: SteamGameServer003`

ghost:
Progress made with the recent 3.16-Beta release! It's now possible to view achievements and use the server browser. However, actually connecting to a server gets the message "This Steam account is being used in another location" in the console, along with the message box below: 
![image](https://user-images.githubusercontent.com/3085852/52903462-c85b1a00-31eb-11e9-84ed-347bde87dd0d.png)
However, if one uses `sv_lan 1` it is finally possible to launch a listen server :)

[...]

PR #8375, opened 2025-01-06
Fix up CGameID param on more old Steam interface versions

psychonic:
This addresses most or all of https://github.com/ValveSoftware/Proton/issues/2249, applying the same fix for lsteamclient's InitiateGameConnection in 008 of ISteamUser to the 005, 006, 007, and 009 versions of the interface. Disassembly shows the same compiler optimization is present on these as well, which is consistent with the issues reported. Newer versions of this interface dropped the CGameID parameter and are implicitly unaffected. Version 004 has a CGameID parameter, but it doesn't appear to perform the same dereferencing. Without the fix, calls to `ISteamUser::InitiateGameConnection` on the affected interface versions most frequently lead to a crash, or at best, authentication will just fail, quickly leading to a disconnect. In testing with the fix, a crash no longer occurred on server connection in NEOTOKYO (Steam app 244630), a Source SDK 2006 mod. The crash is reproducible on Proton 9.0-4.

Edit: Per my findings on that same issue, [here](https://github.com/ValveSoftware/Proton/issues/2249#issuecomment-2574145219) - The same fix is also applied to all of the ISteamUsersStats001 and 002 interface functions. I admit that I didn't manually check every function on every version of those two very old (pre-1.00) interface versions, but they all have a first parameter of CGameID, and the 12 or so function versions I didn't at did have the dereferencing issue present. In testing, this fixed the startup crash reported in that issue with the mod version of Black Mesa on Source SDK 2007. I also confirmed that the achievements panel opened and displayed properly.

This is my first contribution to this repo, so please let me know if I missed anything required. I didn't see any contributing guidelines.

For reference, https://github.com/ValveSoftware/Proton/issues/2249#issuecomment-2574115896

> 
> Part of the original issue reported here, creating/joining servers on Source 2007 mods, was fixed a couple of years ago, in cd60b70
> 
> I'm not sure about the achievements issue. There may still be a problem there, judging by ISteamUserStats crash still being reported in #2249 (comment) earlier this year.
> 
> There are other reports in this thread citing the same or similar issues in Source 2006 mods. Part of that is because the server connect fix for the Source 2007 mods was only applied for the specific version of the ISteamUser interface that they target. However, the version targeted by Source 2006 mods, as well as a couple of other versions of the interface are also affected by the same issue.

### DIFF
--- a/lsteamclient/cppISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001.cpp
+++ b/lsteamclient/cppISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001.cpp
@@ -9,7 +9,7 @@ NTSTATUS ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001_GetNumStats( void *
 {
     struct ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001_GetNumStats_params *params = (struct ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001_GetNumStats_params *)args;
     struct u_ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001 *iface = (struct u_ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001 *)params->linux_side;
-    params->_ret = iface->GetNumStats( params->nGameID );
+    params->_ret = iface->GetNumStats( &params->nGameID );
     return 0;
 }
 
@@ -17,7 +17,7 @@ NTSTATUS ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001_GetStatName( void *
 {
     struct ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001_GetStatName_params *params = (struct ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001_GetStatName_params *)args;
     struct u_ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001 *iface = (struct u_ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION001 *)params->linux_side;
-    params->_ret = iface->GetStatName( params->nGameID, params->iStat );
+    params->_ret = iface->GetStatName( &params->nGameID, params->iStat );
     return 0;
 }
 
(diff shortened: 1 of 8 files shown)


gofman:
Thanks for the patch! Some of the fixup were already added by now, but indeed more ISteamUser interfaces can use that. Now I looked up all the remaining cases of passing CGameID and it looks like the fixup is just always needed, regardless of interface version or method. So I pushed a bit different patch (currently in Proton Experimental bleeding edge branch): https://github.com/ValveSoftware/Proton/commit/9cab9ca10876bf759de94b8a314cda2b76df24f2 
 
```
commit 9cab9ca10876bf759de94b8a314cda2b76df24f2
Author: Paul Gofman <pgofman@codeweavers.com>
Date:   Tue Sep 9 10:55:27 2025 -0600

    lsteamclient: Always pass CGameID by reference.
    
    Based on a patch by Nicholas Hastings <nshastings@gmail.com>.
    Link: https://github.com/ValveSoftware/Proton/pull/8375
    
    CW-Bug-Id: #25907
```

psychonic:
@gofman  Thanks for looking at that and following up

I believe that will handle fixing most of the issues in https://github.com/ValveSoftware/Proton/issues/2249, which is great

psychonic:
Actually, @gofman , I checked with a user that was helping me test my changes, and they claim that the auth piece is still crashing on the bleeding edge branch, but now on connection rather than shortly after. This can be reproduced in app `244630` (NEOTOKYO, free Source SDK 2006 mod), by attempting to connect to a game server.

I'm not sure if this is a result of the difference in the generated code of your patch and mine, or if there is an unrelated change causing that crash now.

When I last tested a build with my changes here, the client was able to connect to a server and maintain connected without crashing.

gofman:
Could you please:
1. Attach default Proton log (recorded with ```PROTON_LOG=1 %command%``` launch parameters).
2. Describe the exact steps how to reproduce the issue.

UPDATE: if repro is convoluted we may start from p. 1., see the log and if it will be all clear from it the repro won't be needed.

mvoolt:
> Could you please:
> 
> 1. Attach default Proton log (recorded with `PROTON_LOG=1 %command%` launch parameters).
> 2. Describe the exact steps how to reproduce the issue.
> 
> UPDATE: if repro is convoluted we may start from p. 1., see the log and if it will be all clear from it the repro won't be needed.

I'm the said user, apologises for the inconvenience but I actually had a dll hook plugin installed which is meant to fix this (but crashes Proton Experimental, this is simply an user-made error), I just tested bleeding-edge properly now and I am not getting any VAC disconnects now :tada:

psychonic:
Thanks for confirming. It sounds like we're good now

PR #8443, opened 2025-02-08
Properly detect whether podman or docker should be used.

GloriousEggroll:
Fedora ships podman with a podman-docker package that uses a bash script for /usr/bin/docker which runs `exec /usr/bin/podman @`, This does not work correctly with the current check mechanism in proton's configure script, resulting in invalid permission settings:
```
!! File owner's UID doesn't map to 0 or 1000 in the container.
!! Don't know how to map permissions. Please check your docker setup.
```
By checking if the /usr/bin/docker file contains contains 'bin/podman' we can easily determine whether or not to actually use podman before any other checks are done.


ivyl:
Inspecting files like this feels fairly brittle. I think we can flip the order of the checks instead and assume that if a `podman` is there and is working that's what the person wants to use over `docker`.

PR #8560, opened 2025-03-25
Fixed spelling and grammatical errors

rbrasga:
Not sure if this is worth the effort, but I fixed a lot of typos.

I manually reviewed each line change for accuracy.


ivyl:
openvr and steamworks headers are direct source imports from other projects,see: https://github.com/ValveSoftware/openvr/ and: https://partner.steamgames.com/doc/sdk.

Since those are external imports we want to keep them as close to originals, even if that means typos.

As of other changes, like the ones to the `.md` those are welcome. If you can trim the YPR to only contain those I can merge it.

PR #8659, opened 2025-05-05
Switch blank video from smpte to a missing-codecs message

nirbheek:
/r/linux_gaming and other forums regularly get posts asking about "weird colors" in games with cutscenes:

https://www.reddit.com/r/linux_gaming/comments/1kf59i6/fragpunk_showing_weird_colors/

https://www.reddit.com/r/linux_gaming/comments/1k47643/blue_prince_game_launches_but_when_i_click_new/ 

https://www.reddit.com/r/linux_gaming/comments/1ftevcl/star_wars_jedi_survivor_is_giving_me_a_weird/

Putting a message in there is probably more useful.

Also switch audio to use wave=silence, which is more correct.

This is what the video will look like after this change: 
image. As a bonus, the video will also be smaller.

### DIFF
--- a/media/Makefile
+++ b/media/Makefile
@@ -2,7 +2,7 @@ all: blank.mkv blank.ptna
 
 blank.mkv:
 	#120 frames @ 30 FPS == 4 seconds
-	gst-launch-1.0 videotestsrc num-buffers=120 pattern=smpte ! 'video/x-raw,format=(string)I420,width=320,height=240,framerate=(fraction)30/1' ! av1enc ! queue ! mux. audiotestsrc num-buffers=400 freq=0 samplesperbuffer=441 ! 'audio/x-raw,rate=48000,channels=2' ! opusenc ! queue ! mux. matroskamux name=mux ! filesink location=blank.mkv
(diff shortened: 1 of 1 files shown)


nirbheek:
Note that this change requires the `textoverlay` element which is part of the `pango` plugin. Since this MKV is generated at build-time this doesn't change the runtime requirements of Proton, but it does require that the gstreamer that you use on your build machines ship that `pango` gstreamer plugin. I noticed that this build-env gstreamer is already different from the gstreamer you ship with Proton since it ships with an av1 encoder, and Proton does not.

Note that the `pango` plugin is shipped by default on all Linux distros, so it should be easy to gain access to.

eloj:
Could you overlay it on the test pattern? That may be the best of both worlds.

nirbheek:
> Could you overlay it on the test pattern? That may be the best of both worlds.

Done. I also made the pattern move and made it loop perfectly. I figured that you might've wanted movement to clarify that the game is not frozen. The change was:

```
- num-buffers=120 pattern=black
+ num-buffers=160 horizontal-speed=2
```

nirbheek:
Just a gentle poke about this, since I saw today that another person reported this on /r/linux_gaming: https://www.reddit.com/r/linux_gaming/comments/1mn9ql3/what_in_the_world_is_this/

PR #8747, opened 2025-05-29
configure.sh: Fix default build-name (regression introduced in 82eb0051)

twhitehead:
Have to put setting environment variable on its own line as otherwise it just just sets that value for that one line.


ivyl:
Merged into experimental. It's already in bleeding-edge and from there it will propagate to all the other versions.

Thanks for spotting this!

PR #8337, opened 2024-12-23
Adding support for Banter

PablosCorner:
This pull request includes a small change to the `proton` file. The change adds a new game, "Banter," to the default compatibility configuration list. This allows for the Vuplex in-game browser to work properly under Proton.

* [`proton`](diffhunk://#diff-469dfc0b18c73eb263c4a1476545e7b064a7f862ef53827c22026c97268b2ebfR1194): Added "Banter" (ID: 2564460) to the default compatibility configuration list.

PR #8784, opened 2025-06-09
Add support for Banter which uses an in app browser from vuplex. 

sidequestlegend:
This is instead of the PR into Proton 9: https://github.com/ValveSoftware/Proton/pull/8337


kisak-valve:
Hello @sidequestlegend, please do not intentionally create duplicate pull requests. There's no value in the duplication.

Closing as a duplicate of #8337.

PR #8943, opened 2025-08-01
Fix libraryfolders.vdf index incrementing in setup_steam_files

### LINKED ISSUE #8899: Proton 10.0: setup_steam_files never increments the index, causing programs which parse libraryfolders.vdf to break
Hello,
The Proton 10 version of the setup_steam_files command fails to increment the index of the library folder when writing libraryfolders.vdf, meaning that every single library folder is given index 0. This breaks programs which parse libraryfolders.vdf, such as https://github.com/soulsmods/ModEngine2.
Tested with GE-Proton10-8, but the version of steam.c is identical to the one in the official repo and the bug is still present there as well.

EDIT: I also tested with the official Proton 10.0-1 (beta) release, and the same exact issue is present.

https://github.com/ValveSoftware/Proton/blob/3a269ab9966409b968c8bc8f3e68bd0d2f42aadf/steam_helper/steam.c#L620-L684

The version of steam.cpp used in Proton 9.0 does not have this issue, as it explicitly uses ++index after writing a line to the buffer.
https://github.com/ValveSoftware/Proton/blob/a62e072e949676917c9f7724bcd34b4e2f8bb53d/steam_helper/steam.cpp#L1472-L1587

To test, the data in `$STEAM_COMPAT_DATA_PATH/pfx/drive_c/Program Files (x86)/Steam/steamapps/libraryfolders.vdf` _should_ contain the following:
```
"LibraryFolders"
{
	"0"
	{
		"path"		"Z:\\home\\umgak\\.local\\share\\Steam"
	}
	"1"
	{
		"path"		"Z:\\home\\umgak\\.local\\share\\Steam"
	}
}
```

On affected versions, it instead contains:
```
"LibraryFolders"
{
	"0"
	{
		"path"		"Z:\\home\\umgak\\.local\\share\\Steam"
	}
	"0"
	{
		"path"		"Z:\\home\\umgak\\.local\\share\\Steam"
	}
}
```
Note that the index of the second entry is still 0. This breaks parsers of libraryfolders.vdf and is unexpected behavior based on previous versions and Windows. The fix should be as simple as adding `++index;` after the two `pos += strappend [...]` lines in that function.

killerdevildog:
- Add missing ++index statements after each library folder entry
- Fixes issue where all library folders were assigned index 0
- Resolves compatibility issues with ModEngine2 and other parsers
- Restores behavior consistent with Proton 9.0

Fixes #8899

### DIFF
--- a/steam_helper/steam.c
+++ b/steam_helper/steam.c
@@ -648,6 +648,7 @@ static void setup_steam_files(void)
         else
         {
             pos += strappend( &buf, &buf_size, pos, "\t\"%u\"\n\t{\n\t\t\"path\"\t\t\"%s\"\n\t}\n", index, str );
+            ++index;
             free( str );
         }
     }
@@ -671,6 +672,7 @@ static void setup_steam_files(void)
         else
         {
             pos += strappend( &buf, &buf_size, pos, "\t\"%u\"\n\t{\n\t\t\"path\"\t\t\"%s\"\n\t}\n", index, str );
+            ++index;
             free( str );
         }
         free( path );


gofman:
I've edited patch description and pulled the patch to Proton (currently in Proton Experimental bleeding-edge branch): https://github.com/ValveSoftware/Proton/commit/7c09f006c1cb64b7de508fc5d01f2137e45c70d0

Thanks!

Issue #2447, opened 2019-03-23
Neotokyo (244630)


glubsy:
# Compatibility Report
- Name of the game with compatibility issues: Neotokyo (source mod, directx9)
- Steam AppID of the game: 244630

## System Information
- GPU: GeForce GTX 670MX
- Driver/LLVM version: nvidia 418.43
- Kernel version: 4.19.29-1-lts (Arch Linux)
- Link to full system information report: [gist](https://gist.github.com/glubsy/8930de18e83adb497905097091e9789c/97b83592b54a2188abee300561375badeb100670)
- Proton version: 3.16.8 Beta

## I confirm:
- that I haven't found an existing compatibility report for this game.
- that I have checked whether there are updates for my system available.

with PROTON_LOG=1 %command%: [log](https://gist.github.com/glubsy/e84d14e9fbb8d76450887779b95b7dbd)

Also tried with PROTON_USE_WINED3D=1 (same log above)

## Symptoms 
Game crashes as soon as it reaches the main menu. 

## Reproduction
Start the game with no modification. 

Tried gdb_run from the /tmp directory and noticed this after typing "continue" (may or may not be related):
```
00000026:00000027: output debug string ("AppFramework : Unable to load module Z:\\ssd_data\\SteamLibrary\\steamapps\\common\\NEOTOKYO\\bin\\filesystem_steam.dll!\n")
00000026:00000027: output debug string ("Unable to load Z:\\ssd_data\\SteamLibrary\\steamapps\\common\\NEOTOKYO\\bin\\filesystem_steam.dll")
0026:0027: loads DLL  @0x7d5d0000 (0<0>)

Program received signal SIGFPE, Arithmetic exception.
0x7ef19f32 in floorf64 () from /usr/lib32/libm.so.6
```

kisak-valve:
Hello @glubsy, your log is missing a header which was added with Proton 3.16-7, which hints that you're using an unexpected version of Proton. Please opt into Steam's beta client and/or go to Steam -> Library dropdown -> Tools and install Proton 3.16 beta from the list.

`err:steamclient:create_win_interface Don't recognize interface name: SteamClient006` from the log suggests Proton 3.16-7 or newer is needed for this game (https://github.com/ValveSoftware/Proton/commit/7008e8bf21f08640e03e33299c6ca42b44ae5928).

glubsy:
[...]

PR #8985, opened 2025-08-20
lsteamclient: Fix crash on ISteamUserStats version 2

### LINKED ISSUE #8979: Ninja Reflex: Steamworks Edition (13000)
# Compatibility Report
- Name of the game with compatibility issues: Ninja Reflex: Steamworks Edition
- Steam AppID of the game: 13000

## System Information
- GPU: GTX 1080
- Video driver version: 550.163.01
- Various system info: https://gist.github.com/Fighter19/1cf04455902519e4aeedb1a1c4dff3e6
- Kernel version: 6.14.0-27-generic
- Proton version: 7.0 to Experimental

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

[steam-13000.log](https://github.com/user-attachments/files/21807151/steam-13000.log)

## Symptoms 
Starting up the game instantly crashes

## Reproduction
Press play

## Analysis
The game uses a custom steam_api.dll using probably an obsolete version of ISteamUserStats.
(strings suggest STEAMUSERSTATS_INTERFACE_VERSION002).

The calls for example to presumably GetAchievement (vtable+0x24) always pass app id and zero as the first two parameters.

Publically available documentation implies that passing these values is unexpected.

Code within the provided function by SteamClient crashes the game, whilst dereferencing the app id, as if it was a pointer.

My gut feeling is, that the wrong steam_api.dll gets loaded.

System info should be unnecessary, but is provided as pro forma.

Fighter19:
Applies the same changes as done in 09846111b.

Might need to be extended for other versions as well.
Fixes #8979 (Ninja Reflex: Steamworks Edition)

### DIFF
--- a/lsteamclient/cppISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002.cpp
+++ b/lsteamclient/cppISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002.cpp
@@ -9,7 +9,7 @@ NTSTATUS ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002_GetNumStats( void *
 {
     struct ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002_GetNumStats_params *params = (struct ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002_GetNumStats_params *)args;
     struct u_ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002 *iface = (struct u_ISteamUserStats_STEAMUSERSTATS_INTERFACE_VERSION002 *)params->u_iface;
-    params->_ret = iface->GetNumStats( params->nGameID );
+    params->_ret = iface->GetNumStats( &params->nGameID );
     return 0;
 }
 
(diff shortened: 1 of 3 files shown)


gofman:
Thanks for the analysis and the patch! Indeed, I checked that it is the same for all the methods of ISteamUserStats (not just RequestCurrentStats)  for interface versions 001, 002. Also, now that Unix side lsteamclient code is C++, there is no need to fixup the method implementation, just using 'CGameID &' instead of 'CGameID *' does the thing.

So I pushed a different patch to Proton which fixes that for the whole ISteamUserStats 001 / 002 (currently in Experimental bleeding edge branch https://github.com/ValveSoftware/Proton/commit/d842e2e3fe441baf6b41a59ab0cc8a5591a257ad)  which however references your contribution.

Fighter19:
I'm glad it could be resolved and that the reference it kept!

mvoolt:
> Thanks for the analysis and the patch! Indeed, I checked that it is the same for all the methods of ISteamUserStats (not just RequestCurrentStats) for interface versions 001, 002. Also, now that Unix side lsteamclient code is C++, there is no need to fixup the method implementation, just using 'CGameID &' instead of 'CGameID *' does the thing.
> 
> So I pushed a different patch to Proton which fixes that for the whole ISteamUserStats 001 / 002 (currently in Experimental bleeding edge branch [d842e2e](https://github.com/ValveSoftware/Proton/commit/d842e2e3fe441baf6b41a59ab0cc8a5591a257ad)) which however references your contribution.

@gofman Any chance you could bring back the fixup back? Games like NEOTOKYO need it. If you need any more information, https://github.com/ValveSoftware/Proton/pull/8375 has it more in detail.

If testing this on NEOTOKYO, you'll specifically need to join a public server (or host a dedicated server with sv_lan 0) and wait 2-3 minutes. The server should disconnect the client with a VAC error

Thanks and hope you get back to this!

kisak-valve:
Hello @mvoolt, add a note to #2447 with the current game behavior and a Proton log of the problem with Proton Experimental [bleeding edge].

gofman:
What do you mean by bring the fixup back, I think I wasn't removing any, only adding? Anyway, thanks for the information. The fixups from the referenced PR seem valid (as well as some other cases which are not listed neither here nor there), I've  pushed a more generic change to Proton Experimental (currently in bleeding edge branch), commented there in https://github.com/ValveSoftware/Proton/pull/8375 .

PR #9361, opened 2026-01-02
proton: Add OPENSSL_ia32cap env variable for Kabounce

rgalhos:
Kabounce is affected by the [OpenSSL* SHA Crash Bug](https://www.intel.com/content/www/us/en/developer/articles/troubleshooting/openssl-sha-crash-bug-requires-application-update.html). Adding the `OPENSSL_ia32cap=~0x20000000` env [fixes the issue](https://steamcommunity.com/app/431930/discussions/2/5400412499751434799/).


kisak-valve:
This should probably be bumped up the list to match the preexisting sorting.

PR #9455, opened 2026-02-01
Refactor: moved list of games appid to appids.py

QtRoS:
Automatically parsed and replaced (with regexp, not AI) _most_ of appid literals in proton. Why? Because 3 year old comment says:

> #hopefully short-lived, app-specific workarounds for Proton bugs

Seems that it's time to make such lists legit, because makes codebase look a bit more solid :)

BTW, I left old ids in comments like this:
```python
if appid in [
                AppID.AS_DUSK_FALLS, # 1341820
                AppID.CREATIVERSE, # 280790
                AppID.THE_ELDER_SCROLLS_ONLINE, # 306130
```

**only** to simplify review process. Once it's done they can be removed IMO.

P.S. Maybe this function worth adding to `appids.py`:

```python
def name_to_constant(name):
    name = name.upper()
    name = re.sub(r"[\"'\[\]\(\):®]", '', name)
    name = re.sub(r'[ &\-\.,/+]+', '_', name)
    name = re.sub(r'__+', '_', name)
    return name.strip('_')
```

Later everyone can use it to add new games.


QtRoS:
Any reviews on this?..

ivyl:
Unnecessary code churn that makes looking up games and their overrides harder.

You can no longer look up just `Tell Me Why 2` but have to remember to use underscores for `TELL_ME_WHY_CHAPTER_2`. Same with looking up via the game id, which we do even more often, - they are now opaquely contained in a separate file for the name-to-appid mapping.

QtRoS:
@ivyl thanks for the review! My point was that game names in comments are not reliable and error-prone. But I guess it's not a problem if you think so. Anyway, feel free to reopen PR if you change your mind. Cheers!

PR #9500, opened 2026-02-15
fix capitalization consistency issues in README.md

FeltMacaroon389:
Various terms were incorrectly or inconsistently capitalized in README.md. Additionally, there was some poorly formatted text. This takes care of that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ locally](#install-proton-locally) section for details.
 deploy Proton to Steam users via Steamworks.
 
 `make module=<module> module` - build both 32- and 64-bit versions of the
-specified wine module. This allows rapid iteration on one module. This target
+specified WINE module. This allows rapid iteration on one module. This target
 is only useful after building Proton.
 
 `make dxvk` / `make vkd3d-proton` - rebuild DXVK / vkd3d-proton.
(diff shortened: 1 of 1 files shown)


Chiitoo:
The capitalisation that upstream developers and documentation use is Wine, which I would consider the more correct form, rather than WINE, even if it seems someawhats popular still.

See for example:

- https://gitlab.winehq.org/wine/wine/-/wikis/FAQ
- https://gitlab.winehq.org/wine/wine/-/wikis/Wine-History

FeltMacaroon389:
Thanks for the comment, I've updated things accordingly.

README.md
```diff
@@ -249,8 +249,8 @@ and [docs/DEBUGGING-WINDOWS.md](docs/DEBUGGING-WINDOWS.md).
 
 For use with [clangd](https://clangd.llvm.org/) LSP server and similar tooling.
 
-Projects built using cmake or meson (e.g. vkd3d-proton) automatically come with
-`compile_commands.json`. For autotools (e.g. wine) you have to [configure the
+Projects built using CMake or Meson (e.g. vkd3d-proton) automatically come with
+`compile_commands.json`. For autotools (e.g. WINE) you have to [configure the
```
SolarWings, line 253:
```suggestion
`compile_commands.json`. For autotools (e.g. Wine) you have to [configure the
```

I noticed you were correcting it from WINE -> Wine. Does this need correction too?

Issue #8896, opened 2025-07-11
Warudo (2079120)


emma-the-rock:
# Compatibility Report
- Name of the game with compatibility issues: Warudo
- Steam AppID of the game: 2079120

## System Information
- GPU: NVIDIA GeForce RTX 4060
- Video driver version: NVIDIA 575.64
- Kernel version: 6.15.5-200.fc42.x86_64
- Link to full system information report as [Gist](https://gist.github.com/): https://gist.github.com/emaxoda/47ba5a6d7052efeb6f5f32849b532318
- Proton version: Proton Experimental

## I confirm:
- [x] that I haven't found an existing compatibility report for this game.
- [x] that I have checked whether there are updates for my system available.

[steam-2079120.tar.gz](https://github.com/user-attachments/files/21173970/steam-2079120.tar.gz)

## Symptoms

App crashes on launch when using anything newer than Proton 8.0-5.

## Reproduction

Use anything newer than Proton 8.0-5 and the app will crash, also it can be fixed by disabling NVAPI (app has DLSS support, although you need a license to have access to it) with the PROTON_DISABLE_NVAPI=1 env var.

emma-the-rock:
Another bug I found is that if you use any proton version newer than Proton 8.0-5 with the NVAPI workaround the app will have a memory leak until the system gets OOM.

[steam-2079120.tar.gz](https://github.com/user-attachments/files/21175052/steam-2079120.tar.gz)

<img width="1122" height="861" alt="Image" src="https://github.com/user-attachments/assets/a7869429-bea7-41df-8509-4379f8f64247" />

As you can see the app is using around 14GB of RAM but started at around 3GB

To reproduce this you only need to use Proton Experimental and 'PROTON_DISABLE_NVAPI=1' as an env var and let the app running in the background.

emma-the-rock:
after extensive testing (two people running the app for more than 48 hours combined), on the proton 10 beta 2 and proton GE 10-10 and using the PROTON_USE_WOW64 env var seems to fix the mem leak, without it the mem leak persist so it probably is a regression on a 32bit library/dll.
@kisak-valve

emma-the-rock:
<img width="1562" height="1404" alt="Image" src="https://github.com/user-attachments/assets/1c1d5541-58b7-45c1-96c3-f7e377373c0d" />
warudo running for more than 24 hours on proton 10 beta 2 with PROTON_USE_WOW64 and PROTON_DISABLE_NVAPI=1

[...]

PR #9588, opened 2026-03-18
Add Warudo to proton disablenvapi

madalee-com:
On Proton 9 and up Warudo will crash on launch with NVIDIA cards unless NVAPI is disabled.

This adds Warudo to the list of exceptions for disablenvapi

### DIFF
--- a/proton
+++ b/proton
@@ -1361,6 +1361,7 @@ def default_compat_config():
                 "435150", #Divinity: Original Sin 2 - Definitive Edition
                 "2176900", #Fablecraft
                 "2853730", #Skull and Bones
+                "2079120", #Warudo
                 ]:
             ret.add("disablenvapi")
 
--- a/proton
+++ b/proton
@@ -1390,6 +1390,7 @@ def default_compat_config():
 
         if appid in [
                 "2698940", #The Crew Motorfest
+                "2079120", #Warudo
                 ]:
             ret.add("hidenvgpu")
 


madalee-com:
Also added Warudo to the hidenvgpu section.  This prevents a "slow" memory leak, around 360MB an hour.

hoshinolina:
Can confirm this is needed.

ivyl:
@Saancreed and @jp7677 - they may want to have a look at the nvapi crash first before we add a hack. The leak may be related to it.

Saancreed:
From what I see in logs attached to https://github.com/ValveSoftware/Proton/issues/8896, NVAPI dll isn't even loaded at all by the application. I suspect this isn't something caused by NVAPI itself, but the application doing _something_ when it sees Nvidia VendorId.

It does, however, attempt to load `nvcuda.dll`, except Proton's default implementation of that library is just a stub that will crash on any attempt to actually use it. I believe Divinity Original Sin 2 suffers from the same issue. Perhaps it would work with a more proper implementation, like SveSop's one from his nvidia-libs project (older version that still supported 32-bit applications).

But for now I'm mostly curious if just `WINEDLLOVERRIDES='nvcuda='` or `DXVK_CONFIG='dxgi.customVendorId=1002'` would be enough for a viable workaround, because it really looks like something out of scope of our little project, even if disabling NVAPI accidentally fixes it.

AdalynBlack:
I did some tests to try and provide more information. All tests were done on Proton Experimental through Steam

No Launch Options: Unity crashes immediately ([Logs](https://github.com/user-attachments/files/27721379/steam-2079120.log))
`WINEDLLOVERRIDES='nvcuda=' %command%`: No Change ([Logs](https://github.com/user-attachments/files/27721408/steam-2079120.log))
`DXVK_CONFIG='dxgi.customVendorId=1002' %command%`: No Change ([Logs](https://github.com/user-attachments/files/27721463/steam-2079120.log))
`PROTON_DISABLE_NVAPI=1 %command%`: Warudo launches successfully ([Logs](https://github.com/user-attachments/files/27721519/steam-2079120.log))

PR #9725, opened 2026-04-29
wineopenxr: Fix session list leak and add helper function

TheRougeRoyal:
- Remove duplicate wineopenxr_strdup, use existing strdupA instead
- Fix memory leak in wine_xrDestroySession: session was never removed from session_list before being freed
- Add is_d3d_session() inline helper for session type checks

### DIFF
--- a/wineopenxr/openxr.c
+++ b/wineopenxr/openxr.c
@@ -84,11 +84,11 @@ substitute_extensions[] =
     {"XR_KHR_win32_convert_performance_counter_time", "XR_KHR_convert_timespec_time", TRUE, TRUE},
 };
 
-static char *wineopenxr_strdup(const char *src)
+static char *strdupA(const char *s)
 {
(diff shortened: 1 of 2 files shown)


ivyl:
The change targets `proton_11.0` but is made directly on top of `proton-9.0-4e` which is 2 generations behind. None of the touched code survived.

Is there a real bug hit by anyone in here? Is there a real need to fix this on EOLed version of Proton? It looks like a code churn for the sake of code churn that only adds noise and creates maintainer burden.


But even then the changes are highly suspicious:

1. Unrelated changes grouped into a single commit.

2. The patch description claims to remove duplicate `wineopenxr_strdup()` which is a lie. Instead it renames the function and does some subtle yet meaningless changes to the implementation.

3. The introduced `is_d3d_session()` that's not even used anywhere.

PR #9823, opened 2026-05-26
proton: Add work-around for Diablo II: Resurrected – Infernal Edition.

tgurr:
Recently bought two copies of the game for two different steam accounts and this **initial** issue appeared on two different machines one running latest Fedora Linux and one Exherbo Linux. Both steam accounts used on these machines have not used battle.net for quite some time and probably haven't been linked before so it might be related to that, especially now that the problem is no longer reproducible for me.

image

Without the hint on https://www.protondb.com/app/2536520 where apparently others are running into this initial issue as well I'd have had no idea how to get the game working. There's also for example https://steamcommunity.com/app/2536520/discussions/0/807975542034834556/ on the steam discussions where the workaround is also [mentioned](https://steamcommunity.com/app/2536520/discussions/0/807975542034834556/?ctp=4#c837249693974055295).

I've also tried to verify my suggested fix by creating a new proton prefix but I couldn't reproduce the initial error anymore as the directory got created without any manual interaction and a file `cookie.bin` placed there now:

image

Proton version used: `1779117445 experimental-11.0-20260518b`

### DIFF
--- a/proton
+++ b/proton
@@ -1106,6 +1106,12 @@ class CompatData:
                         os.makedirs(os.path.dirname(dstvdf), exist_ok=True)
                         shutil.copyfile(srcvdf, dstvdf)
 
+                # Diablo II: Resurrected – Infernal Edition initially needs a ClientSdk directory to not hang on error:
+                # You have not been online in the last 30 days. Please start the game while online to check for any login agreements.
+                if os.environ.get("SteamGameId", 0) == "2536520":
+                    diablo2_clientsdk_dir = "/drive_c/users/steamuser/AppData/Local/Blizzard Entertainment/ClientSdk"
+                    makedirs(self.prefix_dir + diablo2_clientsdk_dir)
+
                 #copy openvr files into place
                 makedirs(self.prefix_dir + "/drive_c/vrclient/bin")
                 try_copy(g_proton.arch_pe_dir("wine", True) + "vrclient.dll", "drive_c/vrclient/bin",


ivyl:
The file gets created by the game upon first launch anyway. It might have been a bigger issue in the past that maybe was resolved by a game update?

The issue seems to be still somewhat present but - I've got it the very first time starting the game. It resolves itself on consecutive launches though and past that I'm not able to reproduce. Even with a completely fresh prefix.

I have log from the first launch that exhibited the issue there's nothing really suspicious in there.

I'll try the same account on different machine. I poke a bit more at it.

tgurr:
> The file gets created by the game upon first launch anyway. It might have been a bigger issue in the past that maybe was resolved by a game update?

The file itself indeed, but apparently not when the directory is not present *and* maybe the game/battle.net *knows* when you already have logged in before, as in the last 30 days, and if yes things work, if not you run into that issue at least initially when there's no cookie from an login to *sync* from. At least that's what I ran into and creating the directory allowed the file to get created and the game to work. I have to say I didn't try to launch the game more than maybe 3 times before checking ProtonDB and manually creating the directory which allowed to game to work in the next run.

> The issue seems to be still somewhat present but - I've got it the very first time starting the game. It resolves itself on consecutive launches though and past that I'm not able to reproduce. Even with a completely fresh prefix.
> I have log from the first launch that exhibited the issue there's nothing really suspicious in there.
> I'll try the same account on different machine. I poke a bit more at it.

That's what I experienced as well which made trying to reproduce the issue impossible for me. Maybe you need an account which didn't have any activity the past 30 days? And I haven't had my accounts steam<->battle.net linked before if that may play into this as well.

Thanks for further poking into this!

ivyl:
After looking at this again I wasn't able to get a consistent reproduction. Across a bunch of new accounts and many devices I've seen it once and it fixed itself on the very next launch.

I've seen reports of Windows users getting this error intermittently as well.

If you know about anyone still getting this error a log with `PROTON_LOG=+file,+steamclient,+reg %command%` would be a huge help.

tgurr:
Thanks! I'll see if it happens again for me once I didn't touch the game for > 30 days, so it can take a while until I can give some feedback again - if it then happens to me again. Additionally I *try* to see if I find someone to test it as well who didn't run it yet but not sure if I can find someone.

rlabrecque:
I'm not sure this is necessary in that you just have to launch via battle.net explicitly once every 30 days? Happens on Windows too if you launch d2r.exe directly without bnet running.

Issue #9746, opened 2026-05-05
do_restore in steampipe_fixups fails


richo:
On `1777366833 experimental-11.0-20260428` installed with bazzite, I can't launch games in steam. Attempting to launch proton via the command line, I get

```
lorgecomput 1 % STEAM_COMPAT_DATA_PATH=~/.proton/ ~/.steam/steam/steamapps/common/Proton\ -\ Experimental/proton run     ~ 
Traceback (most recent call last):                           
  File "/home/richo/.steam/steam/steamapps/common/Proton - Experimental/proton", line 2069, in <module>
    g_proton.do_steampipe_fixups()          
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^                                                                                          
  File "/home/richo/.steam/steam/steamapps/common/Proton - Experimental/proton", line 574, in do_steampipe_fixups          
    result_code = steampipe_fixups.do_restore(self.base_dir, fixups_json)                                                  
  File "/var/home/richo/.local/share/Steam/steamapps/common/Proton - Experimental/steampipe_fixups.py", line 90, in do_resto
re                                                                                                                          
    stat_result = os.lstat(this_file)
FileNotFoundError: [Errno 2] No such file or directory: '/home/richo/.steam/steam/steamapps/common/Proton - Experimental/fil
es/lib/wine/i386-windows/plugplay.exe' 
```

After doing a little research it appears that binary is just a shim, so after touching it, it failed in the same way with the next file in that manifest. I patched `do_restore` to ignore that FileNotFoundError and now games launch as expected.

Without a better understanding of what's going on I'm not sure if the correct behaviour here is to create those files if they don't exist, or not attempt to fix their permissions if they don't. Or maybe my proton install is broken and they should *all* exist. I'm happy to open a PR or diagnose further.

danyalahmed1995:
The crash path seems to be `do_restore()` where every `no_write_paths` entry from `steampipe_fixups.json` still exists. I can reproduce the `FileNotFoundError` with a minimal manifest containing a missing `no_write_paths` entry, while existing entries still get their write bits removed as expected.

I am gonna check the smallest safe fix is to skip missing `no_write_paths` entries during restore, since `empty_dirs` already tolerates existing/ missing state.

ivyl:
> On `1777366833 experimental-11.0-20260428` installed with bazzite

[...]

PR #9834, opened 2026-05-30
steampipe_fixups: skip missing no_write_paths entries

danyalahmed1995:
### Summary

`steampipe_fixups.do_restore()` currently assumes every `no_write_paths` entry from `steampipe_fixups.json` still exists. If one of those files is missing, restore fails with `FileNotFoundError` before Proton can launch.

This skips missing `no_write_paths` entries during restore, while still applying the chmod behavior to files that do exist.

For #9746.

### Testing

* Reproduced the `FileNotFoundError` with a minimal manifest containing a missing `no_write_paths` entry.
* Verified existing `no_write_paths` entries still have write bits removed.
* Verified missing `empty_dirs` behavior is unchanged.
* Ran `python -m py_compile steampipe_fixups.py`.

### DIFF
--- a/steampipe_fixups.py
+++ b/steampipe_fixups.py
@@ -87,7 +87,10 @@ def do_restore(path, manifest):
 
     for file_ in no_write_paths:
         this_file = os.path.join(path, file_)
-        stat_result = os.lstat(this_file)
+        try:
+            stat_result = os.lstat(this_file)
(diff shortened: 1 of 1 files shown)


ivyl:
https://github.com/ValveSoftware/Proton/issues/9746#issuecomment-4590920222

Issue #9660, opened 2026-04-11
steampipe_fixups restore empties default_pfx but wineboot is never called to rebuild it


Nicknamn:
# Bug Report: steampipe_fixups.py restore empties default_pfx but wineboot is never called to rebuild it

## System Information
- **OS:** Ubuntu 24.04.4 LTS (Noble Numbat)
- **Kernel:** 6.17.0-20-generic (HWE)
- **GPU:** NVIDIA GeForce GTX 1080 Ti
- **Driver:** nvidia-driver-570 (570.211.01)
- **Steam client:** 1773426488
- **Proton versions affected:** Proton Experimental, Proton 9.0 (Beta), Proton Hotfix, GE-Proton10-32

## Description

At every game launch, `steampipe_fixups.py restore` is called and empties `files/share/default_pfx/` — removing all DLL files from `drive_c/windows/system32/` and `drive_c/windows/syswow64/`, as well as fonts, registry files, and other prefix contents.

After this cleanup, `make_default_prefix()` in the `proton` script is never called to rebuild `default_pfx` because `missing_default_prefix()` only checks whether the **directory exists**, not whether it contains valid content:

```python
def missing_default_prefix(self):
    return not os.path.isdir(self.default_pfx_dir)

def make_default_prefix(self):
    with self.dist_lock:
        local_env = dict(g_session.env)
        if self.missing_default_prefix():
            local_env["WINEPREFIX"] = self.default_pfx_dir
            local_env["WINEDEBUG"] = "-all"
            g_session.run_proc([self.wine_bin, "wineboot"], local_env)
            g_session.run_proc([self.wineserver_bin, "-w"], local_env)
```

Since `steampipe_fixups.py restore` empties the directory without removing it, `missing_default_prefix()` returns `False` and `wineboot` is never invoked. The result is that `default_pfx` remains empty, and every subsequent game launch fails with:

```
FileNotFoundError: [Errno 2] No such file or directory:
'.../files/share/default_pfx/drive_c/windows/system32/d3d8.dll'
```

## Steps to Reproduce

1. Install any Proton version on Ubuntu 24.04 with kernel 6.17
2. Launch any Windows game using Proton
3. Observe that `steampipe_fixups.py restore` empties `default_pfx`
4. Observe that `wineboot` is never called to rebuild it
5. All subsequent game launches fail silently

## Expected Behavior

[...]

PR #9837, opened 2026-05-30
proton: detect incomplete default_pfx

danyalahmed1995:
### Summary

`missing_default_prefix()` only checked whether the `default_pfx` directory exists. If the directory exists but is empty or incomplete, Proton treats it as valid and skips rebuilding it with `wineboot`.

This updates the check to require a few expected default prefix files that later setup code depends on:

* `system.reg`
* `drive_c/windows/system32/d3d8.dll`
* `drive_c/windows/syswow64/d3d8.dll`

For #9660.

### Testing

* Reproduced the validation failure with an empty/incomplete `default_pfx` directory.
* Verified an absent `default_pfx` is still treated as missing.
* Verified an empty `default_pfx` is now treated as missing.
* Verified a partially-created `default_pfx` without the expected sentinel files is treated as missing.
* Verified a sentinel-valid `default_pfx` is treated as present.
* Ran `python -m py_compile proton steampipe_fixups.py`.
* Ran `git diff --check`.

### DIFF
--- a/proton
+++ b/proton
@@ -578,8 +578,13 @@ class Proton:
                             f.write(new_fixup_mtime + "\n")
 
     def missing_default_prefix(self):
-        '''Check if the default prefix dir is missing. Returns true if missing, false if present'''
-        return not os.path.isdir(self.default_pfx_dir)
+        '''Check if the default prefix dir is missing or incomplete. Returns true if missing, false if present'''
+        required_paths = (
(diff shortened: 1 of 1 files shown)


ivyl:
https://github.com/ValveSoftware/Proton/issues/9660#issuecomment-4591322365

PR #9915, opened 2026-06-29
wineopenxr: Remove wrapped XrSession from session list on destroy.

ImSapphire:
This was causing a use-after-free in xrPollEvent if the runtime reuses a previous XrSession handle when the session is destroyed and recreated.

Fixes Prism3D-based games (Euro Truck Simulator 2, American Truck Simulator) never calling xrBeginSession and running the frame loop if the host runtime reuses the XrSession handle when the engine destroys and recreates its session. Previously the game would ignore the XR_TYPE_EVENT_DATA_SESSION_STATE_CHANGED event since the evt->session pointer from the list was incorrect, we would retrieve the old session pointer.

### DIFF
--- a/wineopenxr/openxr_loader.c
+++ b/wineopenxr/openxr_loader.c
@@ -718,6 +718,12 @@ XrResult WINAPI xrDestroySession(XrSession session) {
     return params.result;
   }
 
+  EnterCriticalSection(&session_list_lock);
+
+  list_remove(&wine_session->entry);
+
+  LeaveCriticalSection(&session_list_lock);
+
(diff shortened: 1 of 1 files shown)


Mr-Zero88:
same fix as this PR: https://github.com/ValveSoftware/Proton/pull/9485

gofman:
Thanks, good catch, I picked that to Experimental bleeding-edge branch; from there it should travel to Proton Experimental proper and eventually to Proton stable.

ImSapphire:
thank you!

PR #10115, opened 2026-09-03
dxvk-igdext: Add submodule and copy igdext64.dll into the prefix

damian-tomczak-intel:
dxvk-igdext (https://github.com/GameTechDev/dxvk-igdext) implements
Intel's D3D11 Graphics Extensions API (UAV overlap, MultiDrawIndirect,
depth bounds) for Wine by forwarding calls to DXVK's own native
implementation of those same extensions. Without it, games that use
this API silently fall back to slower default D3D11 behavior instead
of failing, since these extensions are optional performance paths.

This adds dxvk-igdext as a submodule and builds it via CMake/MinGW
(rules-cmake), installing igdext64.dll to lib/wine/igdext/x86_64-windows.

At runtime, the launcher script copies igdext64.dll into the prefix's
DriverStore.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -75,3 +75,6 @@
 [submodule "libsoup"]
 	path = libsoup
 	url = https://github.com/GNOME/libsoup.git
+[submodule "dxvk-igdext"]
+	path = dxvk-igdext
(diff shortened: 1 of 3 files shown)


ivyl:
merged into bleeding-edge

